PR #4, opened 2025-05-02
Add basic README

zanieb:


MichaReiser:
Thanks. This looks good but we probably have to iterate on the license. I created an issue to do that so that we can go ahead and merge this.

PR #27, opened 2025-05-05
Add example of resetting the submodule

zanieb:


CONTRIBUTING.md
```diff
@@ -57,3 +57,10 @@ git switch -c "sync/ruff-${commit}"
 git add ruff
 git commit -m 'Update ruff submodule to https://github.com/astral-sh/ruff/commit/${commit}'
 ```
+
+To restore the Ruff submodule to a clean-state, reset, then update the submodule:
+
+```
+git -C ruff reset --hard
+git submodule update
```
MichaReiser, line 65: I always use `git submodule update --init` but I don't know what the difference is

zanieb: I presume that initializes it if it wasn't yet

PR #35, opened 2025-05-05
setup cargo-dist and build/publish workflows

Gankra:
This is basically an exact copy-paste of the ruff stuff but with `ruff` => `ty` applied pretty blindly.

This likely needs a few iterations to handle the fact that:
* the actions/checkout needs to be recursive
* every path needs to be offset by an extra `./ruff/` segment

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -12,12 +12,12 @@ on:
       plan:
         required: true
         type: string
-  pull_request:
-    paths:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-docker copy.yml
```diff
@@ -0,0 +1,299 @@
+# Build and publish a Docker image.
```
zanieb: Oopsie?

PR #37, opened 2025-05-05
Enable binary builds

zanieb:


.github/workflows/build-binaries.yml
```diff
@@ -12,18 +12,19 @@ on:
...
-      # - .github/workflows/build-binaries.yml
+      - .github/workflows/build-binaries.yml
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
-permissions: {}
+permissions:
+  contents: read
```
Gankra, line 27: curious, why did this change?

.github/workflows/build-binaries.yml
```diff
@@ -43,11 +44,10 @@ jobs:
         with:
           submodules: recursive
           persist-credentials: false
+          token: ${{ github.token }}
```
Gankra: similarly

.github/workflows/build-binaries.yml
zanieb: Because the repo is private and `actions/checkout` failed without it. It must not be needed for public repos.

.github/workflows/build-binaries.yml
zanieb: This can be removed, this was an attempt to debug the above issue.

PR #40, opened 2025-05-05
Bump Ruff to `f6d8f73ed`

zanieb:
From https://github.com/astral-sh/ruff/pull/17868

Not on `main`, but we can start testing from this commit.


Gankra:
So glad I re-enabled the dist `plan` CI

zanieb:
Yeah really helpful!

zanieb:
Except it's not running now 🤔

PR #46, opened 2025-05-05
Include the raw tag in Docker publish labels

zanieb:
In https://github.com/astral-sh/ty/actions/runs/14848358920/job/41687152688, we've published a Python PEP 440 tag but we are trying to consume a Rust version tag later. We could transform the Rust tag into a Python tag, but since `ty version` will report the Rust version tag for pre-releases (and so will the Git tag), we should just _also_ publish a tag for matching Rust pre-release styling.

Follows #44


zanieb:
Finally, success https://github.com/astral-sh/ty/actions/runs/14848531602/job/41687606862

PR #57, opened 2025-05-06
Basic CI workflow

MichaReiser:


.github/workflows/build-docker.yml
```diff
@@ -20,6 +20,9 @@ on:
 env:
   TY_BASE_IMG: ghcr.io/${{ github.repository_owner }}/ty
 
+permissions:
+  contents: read
```
MichaReiser, line 24: @zanieb does this look correct to you for the docker build job

zanieb: Why is it changing? Are you just adding permissions? This will probably need like `packages: read` too?

zanieb: Note we need `packages: write` for some jobs here

There are also permissions defined at

https://github.com/astral-sh/ty/blob/aeff25fea689632d0a6c2fce0aea27fd2c188d98/dist-workspace.toml#L67

MichaReiser: linters are complaining about allowing all permissions. I see that there's already a `packages: "write"` further down.

MichaReiser: Is there a way for me to verify that the job still runs successfully?

zanieb: I don't think so, the build is tested but the rest can't be tested without a release.

PR #61, opened 2025-05-06
Commit the `uv.lock` and `.python-version` files

zanieb:
These are just in my tree right now, but seem proper to commit.


zanieb:
uv will use a consistent Python version when running scripts, which is helpful. We should also probably use `dependency-groups`, which would make the `uv.lock` relevant. e.g., we could replace the uvx `rooster` invocation such that it actually uses locked dependencies.

zanieb:
(I think it's best practice to use these, not ignore them — and I don't think our use-case is an exception in that regard)

AlexWaygood:
Adding the `.python-version` file makes sense to me. Does the `uv.lock` file make sense? When I'm running scripts from the Ruff repo, I nearly always use `uv run --no-project`, because if I don't then uv tries to install Ruff from source using maturin (which takes forever, because Rust builds are so slow :P )

zanieb:
> Does the uv.lock file make sense? When I'm running scripts from the Ruff repo, I nearly always use uv run --no-project, because if I don't then uv tries to install Ruff from source using maturin (which takes forever, because Rust builds are so slow :P )

Well, I set up cache keys for that already (#39) so the rebuilds are less out of control but still slow. However, for dependency groups, you'd do or `--only-group <name>` and we'll skip building the project:

```
❯ uv run --only-dev  python -c "print('hi')"
hi
```

So... yes, the `uv.lock` makes sense — and we should use dependency groups (at the very least, for Rooster).

AlexWaygood:
OK, sounds good!

zanieb:
Moving rooster over there in #64, oh glorious locked dependencies :D

PR #64, opened 2025-05-06
Move `rooster` into a dependency group for locking

zanieb:
As discussed in #61


pyproject.toml
```diff
@@ -7,6 +7,15 @@ authors = [{ name = "Astral Software Inc.", email = "hey@astral.sh" }]
 requires-python = ">=3.8"
 dependencies = []
 
+[build-system]
+requires = ["maturin>=1.0,<2.0"]
+build-backend = "maturin"
```
zanieb, line 12: Just re-ordered so the tool stuff is after the standards

PR #65, opened 2025-05-06
Add script to update `ty.json` in schemastore

MichaReiser:
## Summary

Adds a script to update ty's schema in the schemastore repository

Closes https://github.com/astral-sh/ty/issues/13

## Test plan

https://github.com/astral-sh/schemastore/pull/new/update-ty-52d16b44bcc060bec0a0381a1ea8052bf4479f14

### DIFF
--- a/scripts/update_schemastore.py
+++ b/scripts/update_schemastore.py
@@ -14,11 +14,8 @@ from subprocess import check_call, check_output
 from tempfile import TemporaryDirectory
 from typing import NamedTuple, assert_never
 
-ty_repo = "https://github.com/astral-sh/ty"
-root = Path(
-    check_output(["git", "rev-parse", "--show-toplevel"], text=True).strip(),
-)
-ty_json = Path("schemas/json/ty.json")
+TY_REPO = "https://github.com/astral-sh/ty"
+TY_JSON = Path("schemas/json/ty.json")
 
 
 class SchemastoreRepos(NamedTuple):
@@ -47,7 +44,7 @@ class GitProtocol(enum.Enum):
 
 
 def update_schemastore(
-    schemastore_path: Path, schemastore_repos: SchemastoreRepos
+    schemastore_path: Path, schemastore_repos: SchemastoreRepos, root: Path
 ) -> None:
     if not schemastore_path.is_dir():
         check_call(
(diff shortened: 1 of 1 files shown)


zanieb:
> Check that the ruff submodule has no changes or automatically update it

You could crib this from the `release.sh` script (and / or move it to a `check-submodule.sh` script). I'm unsure it should be updated though.

MichaReiser:
> You could crib this from the release.sh script (and / or move it to a check-submodule.sh script). I'm unsure it should be updated though.

Yeah, updating feels a bit aggressive but I think it's important to warn if someone is on a stale revision. I've to figure out how to check that.

scripts/update_schemastore.py
MichaReiser, line 1: This is mostly just copy paste from the ruff repository. The only real addition is in `main` where we check the state of the `ruff` submodule

scripts/update_schemastore.py
```diff
@@ -0,0 +1,184 @@
...
+from __future__ import annotations
+
+import enum
+import json
+from pathlib import Path
+from subprocess import check_call, check_output
+from tempfile import TemporaryDirectory
+from typing import NamedTuple, assert_never
+
+ty_repo = "https://github.com/astral-sh/ty"
```
zanieb: It's more idiomatic to use ALL_CAPS for these module level constants.

scripts/update_schemastore.py
```diff
@@ -0,0 +1,184 @@
...
+root = Path(
+    check_output(["git", "rev-parse", "--show-toplevel"], text=True).strip(),
```
zanieb: It's really quite naughty to run a subprocess on import. Could you avoid that? :)

zanieb: Is this just `Path(__file__) / ..`?

scripts/update_schemastore.py
```diff
@@ -0,0 +1,184 @@
...
+)
+ty_json = Path("schemas/json/ty.json")
```
zanieb: This should be relative to `Path(__file__)` which makes the script working directory agnostic.

MichaReiser: This is a relative path to the schemastore repository and it gets concatenated with the path to the schemastore repositroy. So I think what's here is correct.

scripts/update_schemastore.py
MichaReiser: Yes. This was already present in the existing script. I don't mind it using `git` here, considering how much other git operations its running anyway.

MichaReiser: > It's really quite naughty to run a subprocess on import. Could you avoid that? :)

I don't think it matters. This is a one of script. Don't import it!

PR #66, opened 2025-05-07
Add pyproject metadata

zanieb:


pyproject.toml
```diff
@@ -1,11 +1,41 @@
...
+  "Programming Language :: Rust",
+  "Topic :: Software Development :: Quality Assurance",
+  "Topic :: Software Development :: Testing",
+  "Topic :: Software Development :: Libraries",
+  "Topic :: Software Development :: Python Modules",
+]
+
+[project.urls]
+Repository = "https://github.com/astral-sh/ty"
+Documentation = "https://docs.astral.sh/ty"
```
dhruvmanila: This doesn't exists yet. I think it's fine to include it but just mentioning it.

MichaReiser:
```suggestion
Repository = "https://github.com/astral-sh/ty"
Documentation = "https://github.com/astral-sh/ty"
```


We can change this once we add proper documentation. I don't expect this to happen before the beta

zanieb: Ah, thanks. We should drop it for now.
```suggestion
```

zanieb: (I'll just remove this for now)

MichaReiser: I'd prefer keeping it beause we'll have documentation in the readme

PR #263, opened 2025-05-08
Add uv lock --check to pre-commit

### LINKED ISSUE #253: Add `uv lock --check` to pre-commit

paul-nameless:
Fixes #253


.pre-commit-config.yaml
```diff
@@ -7,6 +7,13 @@ exclude: |
   )$
 
 repos:
+  - repo: local
```
MichaReiser: Could we use https://github.com/astral-sh/uv-pre-commit instead which ensures that `uv` gets installed if it isn't present on the system

PR #292, opened 2025-05-09
Reference documentation

MichaReiser:
## Summary

Extend the release script to copy the reference documentation to the `docs` folder because GitHub doesn't support relative links into submodules (and I think it's confusing if we redirect users to the ruff repository)

Closes https://github.com/astral-sh/ty/issues/304

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -48,3 +48,9 @@ uv run --only-group release \
 
 echo "Updating lockfile..."
 uv lock
+
+echo "Copying documentation"
+cp ruff/crates/ty/docs/cli.md ./docs
+cp ruff/crates/ty/docs/configuration.md ./docs
+cp ./ruff/crates/ty/docs/rules.md ./docs
+git add ./docs
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,6 +4,7 @@ exclude: |
   (?x)^(
     .github/workflows/release.yml|
     ruff/.*|
+    docs/(cli|configuration|rules).md
   )$
 
 repos:
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Start the language server for IDE integration:
 ty server
 ```
 
-For detailed information about command-line options, see the [CLI documentation](ruff/crates/ty/docs/cli.md).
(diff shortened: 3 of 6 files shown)


scripts/release.sh
```diff
@@ -48,3 +48,9 @@ uv run --only-group release \
 
 echo "Updating lockfile..."
 uv lock
+
+echo "Copying documentation"
+cp ruff/crates/ty/docs/cli.md ./docs
+cp ruff/crates/ty/docs/configuration.md ./docs
+cp ./ruff/crates/ty/docs/rules.md ./docs
+git add ./docs
```
MichaReiser, line 56: @zanieb requesting review for this crime ;)

zanieb: Ah gee, that's weird. I'll give that some thought — duplicating it seems rather unfortunate.

zanieb: This is the generated reference documentation? Why aren't we just generating it here?

zanieb: I think it seems reasonable to just move that generation step from the ruff repository to here and require it on submodule update.

MichaReiser: I like it in the ruff repo because it visualizes the impact of changes.

I also don't think it's that different. It's either copying files (which is super fast) or generating them (which is very slow and requires the script to write outside its root)

So I actually prefer this. Seems simpler overall and doesn't require more time

MichaReiser: I'll go and merge this. I understand the concern around duplicating the file in both repositories but I think it's a very minor concern (and comes with trade-offs) and the solution to this is to generate and host our docs outside of github.

zanieb: For me, it's not about the duplication across repositories — it's about the duplication within this repository where, now, the submodule contents could accidentally drift from the documentation. I think it's minor, but it feels like the sloppy approach. Anyway, easy to fix if it causes tangible problems so 🤷‍♀️

MichaReiser: I see, although it's not clear to me how generating the docs in the ty repository would avoid the "risk of divergence". The docs would diverge when updating the ruff module without regenerating the docs. 

However, we could see the divergence as a feature. It gives us versioned documentation that only gets updated with every release ;)

PR #293, opened 2025-05-09
Basic README sections

MichaReiser:
Basic skeleton for the README with some commented out sections. I think this is a good starting point to fill in specific sections.

Most credit goes to the authors of uv's documentation because I did a lot of copy paste :)


carljm (approved):
Thank you!

PR #323, opened 2025-05-12
fix(doc): consistent numbering in CONTRIBUTING.md

samypr100:
## Summary

While reviewing contributing.md, I noticed numbering was a tad off although GH still renders this correctly.


MichaReiser:
This is intentional and [supported by markdown](https://spec.commonmark.org/0.30/#start-number)

Using `1.` for all items has the advantage that it doesn't require changing all numbers when adding a new step (which is very easy to miss)

PR #327, opened 2025-05-12
Add "rules" and "checking a project" sections

MichaReiser:
## Summary

Add documentation about what rules are and how to type check a project

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ uv run ty check src/main.py
 
 ### First-party modules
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -263,3 +335,7 @@ conditions.
     <img src="https://raw.githubusercontent.com/astral-sh/uv/main/assets/svg/Astral.svg" alt="Made by Astral">
   </a>
 </div>
+
+[commands]: https://github.com/astral-sh/ty/blob/main/docs/cli.md
```
MichaReiser, line 370: I switched to using full URls because pypi doesn't support relative links

zanieb: We should probably restore the PyPI readme transform script so these links work properly on branches in PRs, but this seems fine to start

MichaReiser: Doesn't it just replace badges/images? But yes, this is something we could add to it as well

zanieb: It looks like in uv the relative link is turned into a tagged link? (from looking at PyPI)

MichaReiser: that's cool!

PR #329, opened 2025-05-12
Rewrite the getting started documentation

zanieb:
[Rendered](https://github.com/astral-sh/ty/tree/zb/getting-started?tab=readme-ov-file#getting-started)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,87 +11,38 @@ and feature-complete.
 
 ## Getting started
 
-Use [uvx](https://docs.astral.sh/uv/guides/tools/#running-tools) to get started quickly:
+Run ty with [uvx](https://docs.astral.sh/uv/guides/tools/#running-tools) to get started quickly:
+
+```shell
+uvx ty
+```
+
+For other ways to install ty, see the [installation](./docs/README.md#installation) documentation.
+
+If you do not provide a subcommand, ty will list available commands — for detailed information about
+command-line options, see the [CLI reference](./docs/reference/cli.md).
+
+Use the `check` command to run the type checker:
 
 ```shell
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -11,47 +11,72 @@ and feature-complete.
...
 ```
 
-You should see a help menu listing the available commands.
-
-For detailed information about command-line options, see the [commands] reference.
+You should see a help menu listing the available commands. For detailed information about command-line options, see the [commands] reference.
 
 ## Checking a project
 
-The easiest way to type check your project is by running ty through [uv](https://docs.astral.sh/uv/). To do that, add ty as a project dependency:
```
MichaReiser, line 48: This was sort of intentional to use `uv run` over `uvx` here because, as far as I understand, uvx doesn't set the virtual environment automatically. Meaning, your out-of-the-box experience is worse.

zanieb: I'm exploring that, but the experience should be about equivalent with a `.venv` because ty detects that? `uv run` just implements discovery of `.venv` in parents, so I don't think the experience should be much different.

Since it's alpha, I think we might want to encourage the ephemeral experience over "make this a dependency of the project for other developers".

I'm just trying things out here right now though. I'm not sure what the best way to capture all the invocation / installation methods is yet.

MichaReiser: I'm fine either way. I only wanted to point out why I went with `uv run` over `uvx`

zanieb: I think I'm going to prefer `uvx` here because I don't want to encourage people to add it as a dependency (yet) and after a few iterations I think it's cleaner to avoid multiple installation methods in the README itself.

PR #331, opened 2025-05-12
Split the documentation out of the top-level README

zanieb:
Moves most of the documentation into a dedicated `docs/README` to make it easier to reach important content in the top-level readme like "getting involved".

There are some minor changes to the heading levels here and link changes, but otherwise the content is the same. Don't go deep on the getting started section, that's changing in #329


.pre-commit-config.yaml
```diff
@@ -38,11 +38,19 @@ repos:
         additional_dependencies:
           - mdformat-mkdocs==4.0.0
           - mdformat-footnote==0.1.1
+        exclude: |
+          (?x)^(
+            | docs/reference/\w+\.md
+          )$
```
AlexWaygood: should we exclude these from all pre-commit hooks if they're generated files? If so, I'd add them to the global `exclude` regex at the top of this file rather than hook-specific `exclude` regexes

zanieb: I just copy pasted from Ruff's config

zanieb: Oh, that's how it was passing before... these were in the top-level one

zanieb: Thanks!

MichaReiser (approved):

PR #335, opened 2025-05-12
Edit the module and editor sections of the docs

zanieb:


docs/README.md
```diff
@@ -50,7 +50,14 @@ pip install ty
...
 ### First-party modules
 
-First-party modules are Python files that are part of your own project’s codebase, unlike modules from the standard library modules or third-party packages installed via a package manager like uv or pip. By default, ty searches for first-party modules in the project's root folder or the `src` folder (if present). If your project uses a different layout, configure the project's [`src.root`](./reference/configuration.md#root) in your `pyproject.toml`. For example, if your project's code is in the `app` directory, like so:
+First-party modules are Python files that are part of your project source code.
+
+By default, ty searches for first-party modules in the project's root directory or the `src`
+directory, if present.
+
+If your project uses a different layout, configure the project's
+[`src.root`](./reference/configuration.md#root) in your `pyproject.toml`. For example, if your
```
AlexWaygood:
```suggestion
[`src.root`](./reference/configuration.md#root) in your `pyproject.toml` or `ty.toml`. For example, if your
```

docs/README.md
```diff
@@ -50,7 +50,14 @@ pip install ty
...
+project's code is in the `app` directory, like so:
```
AlexWaygood:
```suggestion
project's code is in an `app/` directory:
```

docs/README.md
```diff
@@ -70,41 +77,85 @@ root = "./app"
 
 ### Third-party modules
 
-Third-party modules are external Python packages that are not part of the standard library or your own project’s code. These are typically installed using a package manager like uv or pip and include libraries such as `requests`, `numpy`, or `django`.
+Third-party modules are Python packages that are not part of your project or the standard library.
+These are usually dependencies of your project, and are installed using a package manager like uv or
+pip. Common examples include libraries such as `requests`, `numpy`, or `django`.
```
AlexWaygood:
```suggestion
These are usually declared as project dependencies in a `pyproject.toml` file or similar,
and are installed using a package manager like uv or pip. Examples of popular third-party
libraries are `requests`, `numpy` and `django`.
```

docs/README.md
```diff
@@ -70,41 +77,85 @@ root = "./app"
...
+
+First, ty checks for an active environment using the `VIRTUAL_ENV` environment variable. If not set,
+ty will search for a `.venv` folder in the project root or working directory.
 
-ty searches third-party modules in your project's virtual environment. By default, it looks for a virtual environment in a `.venv` folder located at the project root. If the `VIRTUAL_ENV` environment variable is set, ty will use the path specified there instead. If your project uses a different location for your virtual environment, specify the location by setting the [`environment.python`](./reference/configuration.md#python) configuration or [`--python`](./reference/cli.md#ty-check--python) CLI option.
+The Python environment may be explicitly specified using the
+[`environment.python`](./reference/configuration.md#python) setting or
+[`--python`](./reference/cli.md#ty-check--python) flag.
+
+When setting the environment explicitly, non-virtual environments can be provided.
```
AlexWaygood: I find the order in which the information is presented here a little confusing because you mention `--python` after `VIRTUAL_ENV` and `.venv`, but the first thing ty checks is whether it was set explicitly using `--python` or the `environment.python` config option. I think it would be easier to follow if we listed the fallbacks in the order ty considers them: first detail the explicit ways of configuring this, then detail the fallbacks it uses.

I think it's also important to state that the `VIRTUAL_ENV` variable will be implicitly set:
- if you're using `uv run` or an equivalent from pdm/poetry/hatch/etc.
- if you explicitly activate the virtual environment in your shell

docs/README.md
```diff
@@ -70,41 +77,85 @@ root = "./app"
...
 
 ## Python version
 
-The supported Python syntax and standard library functions differ between Python versions. For example, Python 3.10 introduced support for `match` statements and the `sys.stdlib_module_names` symbol. While these features enhance the language, using them in a project targeting an older Python version can lead to compatibility issues.
-ty helps you avoid such compatibility issues by checking your code against the Python version your project targets, and flagging any use of features or standard library symbols not available in that version.
+The Python version affects allowed syntax, type definitions of the standard library, and type
+definitions of first- and third-party modules that are conditional on the Python version.
+
+For example, Python 3.10 introduced support for `match` statements and added the the
```
AlexWaygood:
```suggestion
For example, Python 3.10 introduced support for `match` statements and added the
```

docs/README.md
```diff
@@ -70,41 +77,85 @@ root = "./app"
...
+`sys.stdlib_module_names` symbol. However, if your project also supports Python 3.9, you cannot use
+these new features.
```
AlexWaygood:
````suggestion
`sys.stdlib_module_names` symbol to the standard library. However, if your project also supports
Python 3.9, you cannot use these new features unless they are inside an `if sys.version_info >= (3, 10)` branch:

```python
import sys

if sys.version_info >= (3, 10):
    print(sys.stdlib_module_names)  # okay whatever your python-version is set to, because of the `sys.version_info` condition
else:
    print(sys.stdlib_module_names)  # ty will emit an error here if the configured `python-version` is <3.10
```
````

File: docs/README.md [hunk 7]
```diff
@@ -70,41 +77,85 @@ root = "./app"
...
+which is currently 3.13.
 
-- The [`python-version`](./reference/configuration.md#python-version) value in your configuration or the [`--python-version`](./reference/cli.md#ty-check--python-version) command line option.
-- The lower bound of the project's [`requires-python`](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires) field in your `pyproject.toml`.
-- If neither is specified, ty defaults to the latest stable Python version it supports (Python 3.13)
+ty will not infer the Python version from the Python environment at this time.
+
+The Python version may be explicitly specified using the
+[`python-version`](./reference/configuration.md#python-version) setting or the
+[`--python-version`](./reference/cli.md#ty-check--python-version) flag.
```
AlexWaygood, line 138: it's a bit weird to me that we mention these last, since they're the things that ty checks for first

docs/README.md
```diff
@@ -70,41 +77,85 @@ root = "./app"
...
+To disable this functionality, set [`respect-ignore-files`](./reference/configuration.md#respect-ignore-files) to `false`.
 
-Alternatively, you can explicitly pass the paths that ty should check, like so: `ty check src scripts/benchmark.py`. We plan on adding dedicated options for including and excluding files in an upcoming release.
+You may also explicitly pass the paths that ty should check, e.g.:
+
+```shell
+ty check src scripts/benchmark.py
+```
+
+We plan on adding dedicated options for including and excluding files in an upcoming release.
```
AlexWaygood:
```suggestion
We plan on adding dedicated options for including and excluding files in future releases.
```

AlexWaygood (approved):
this is excellent, thanks

docs/README.md
zanieb: I think the existing structure of covering the default behavior (which is what most people will use) makes sense before discussing the fact that you can override that behavior? I am generally hesitant to cover the explicit options first here.

> I think it's also important to state that the VIRTUAL_ENV variable will be implicitly set:

This makes sense to me.

docs/README.md
zanieb: I'm hesitant to go into more detail here, it seems like a bit of a distraction and "or similar" is pretty load bearing. What's they key point we're trying to establish here?

AlexWaygood: I wasn't really sure what "these are usually dependencies of your project" really meant in your current text. It feels almost tautological -- third-party modules are dependencies, yes, but what's the definition of a dependency? It's a third-party module.

zanieb: Haha see, I think the point is to provide that sort of tautological "third-party modules is just a fancy word for dependencies" statement since we're explaining to a beginner what we mean by this concept. I expect people to know what a "dependency" is. I can play with this language though...

AlexWaygood: > I expect people to know what a "dependency" is

Ah, I do not!! I've seen lots of Python noobs who've been forced to type checkers by the company they're working for, and who know very little at all about Python packaging 😄

zanieb:
```suggestion
These are usually declared as dependencies in a `pyproject.toml` or `requirements.txt` file
and installed using a package manager like uv or pip. Examples of popular third-party
modules are `requests`, `numpy` and `django`.
```

docs/README.md
AlexWaygood: What about something like this:

```md
The Python environment ty should use can be explicitly specified, but ty can also automatically
infer your project's environment in many cases.

ty only supports _automatic_ discovery of virtual environments at this time.

When discovering virtual environments, ty first checks for an active environment using the `VIRTUAL_ENV`
environment variable. If not set, ty will search for a `.venv` folder in the project root or
working directory.

The Python environment may be explicitly specified using the
[`environment.python`](./reference/configuration.md#python) setting or
[`--python`](./reference/cli.md#ty-check--python) flag. When setting
the environment explicitly, non-virtual environments can be provided.
```

So, we still detail the automatic discovery before the explicit config options, but we make it slightly clearer upfront that explicit config is supported and non-virtual environments are supported.

docs/README.md
zanieb: That's sort of not the target audience for _this_ documentation, but anyway I'm fine with changing it.

AlexWaygood: that LGTM!

docs/README.md
zanieb: I don't love it :D

zanieb: I'm trying to rephrase it for you, but this pattern is used in each section haha

zanieb: I think if we wanted them to be used by most users, I'd list them first, but the idea is that you don't need the setting.

zanieb: In the same way that we don't lead with "here's how to use `src.root`"

AlexWaygood: yeah I get that... maybe something like this...?

```md
In many cases, ty can also automatically infer your project's environment.
ty only supports _automatic_ discovery of virtual environments at this time.

When discovering virtual environments, ty first checks for an active environment using the `VIRTUAL_ENV`
environment variable. If not set, ty will search for a `.venv` folder in the project root or
working directory.

The Python environment can also be explicitly specified using the
[`environment.python`](./reference/configuration.md#python) setting or
[`--python`](./reference/cli.md#ty-check--python) flag. If either of these
are set, they will take precedence over ty's automatic environment discovery.

When setting the environment explicitly, non-virtual environments can be provided.

Comment (zanieb) [reply docs/README.md]: I pushed a tweak, maybe it's a bit better?

I'm trying to use the same language as the rest of the sections, e.g., "By default, ..."

Comment (AlexWaygood) [reply docs/README.md]: Yeah, the new language looks great, thanks!

File: docs/README.md [hunk 9]
```diff
@@ -70,41 +77,99 @@ root = "./app"
...
+
+### Python environment
+
+The Python environment is used for discovery of third-party modules.
+
+By default, ty will attempt to discover a virtual environment.
+
+First, ty checks for an active virtual environment using the `VIRTUAL_ENV` environment variable. If
+not set, ty will search for a `.venv` directory in the project root or working directory. ty only
+supports discovery of virtual environments at this time.
```
AlexWaygood, line 95: think it would still be great to mention that this is set by uv/poetry/hatch/pdm if they're running `uv run` or similar, and that it's set if you activate the environment explicitly in the shell

docs/README.md
```diff
@@ -70,41 +77,99 @@ root = "./app"
...
 
-ty searches third-party modules in your project's virtual environment. By default, it looks for a virtual environment in a `.venv` folder located at the project root. If the `VIRTUAL_ENV` environment variable is set, ty will use the path specified there instead. If your project uses a different location for your virtual environment, specify the location by setting the [`environment.python`](./reference/configuration.md#python) configuration or [`--python`](./reference/cli.md#ty-check--python) CLI option.
+The Python environment may be explicitly configured using the
+[`environment.python`](./reference/configuration.md#python) setting or
+[`--python`](./reference/cli.md#ty-check--python) flag.
```
AlexWaygood, line 104:
```suggestion
The Python environment may be explicitly configured using the
[`environment.python`](./reference/configuration.md#python) setting or
[`--python`](./reference/cli.md#ty-check--python) flag. This takes
precedence over the virtual environment discovery detailed above.
```

AlexWaygood (approved):

docs/README.md
zanieb: Thanks for your lenience :) haha

docs/README.md
zanieb: Yeah working on that 👍

docs/README.md
zanieb: I don't want to add that :p explicit configuration _always_ takes place over defaults. That's the case for all of our options.

AlexWaygood: eh, fair enough

docs/README.md
zanieb: I do think this is a little awkward to add though. As before, it's a different target audience than the alpha documentation is for.

zanieb: i.e., https://github.com/astral-sh/ty/pull/335/commits/c97679219ab16ef725e3a0034f58e8b3a2996b71

zanieb: I think it sort of interrupts the flow more than it helps, but I don't feel strongly.

AlexWaygood: I feel pretty strongly that this is important information -- it's part of the theme we're trying to describe here, which is "for most cases, you should (hopefully) not have to provide any configuration at all". If everybody starts setting `VIRTUAL_ENV` environment variables because they don't realise that uv is already going to set it for them (or, if they're using pip, because they don't realise it's already set when they activate the environment), that feels like a bad outcome? 😃

The note you added looks great; I'm happy with the text now!

zanieb: I agree it's important, but I think this is the wrong place for it. Those users need "guide"-style documentation that is explicitly targeting beginners (like we have in uv). This documentation _isn't_ for beginners. It's for people trying an experimental type checker. Catering to two audiences like that is just asking for confusing documentation. Anyway 🤷‍♀️

AlexWaygood: that's all fair!

PR #343, opened 2025-05-13
Add reference for editor settings

dhruvmanila:
## Summary

An alternative way to enumerate language server / VS Code settings instead of mentioning them in the extension README (similar to what is done for Ruff via https://docs.astral.sh/ruff/editors/settings/):


docs/editors/settings.md
```diff
@@ -0,0 +1,138 @@
+# Settings
```
zanieb: Should this be `reference/server.md` instead? It'd be nice to keep the reference docs together.

MichaReiser:
Thank you

PR #397, opened 2025-05-14
Use Cargo-style versions in the changelog

zanieb:
Uses https://github.com/zanieb/rooster/pull/70, which should fix dist's changelog -> GitHub Release propagation.


sharkdp:
Thank you for looking into this!

Issue #387, opened 2025-05-14
Add a CI job that fails with a nice error message if generated files are not up to date


AlexWaygood:
We should add a CI job that will fail on PRs like https://github.com/astral-sh/ty/pull/383 so that we can easily tell which PRs touch generated code (and therefore that we should not approve and merge). In this case I remembered that this might be generated code when I saw the PR but I'm worried I won't remember next time (the PR looked correct!). Something like what we have at https://github.com/astral-sh/ruff/blob/cf70c7863c2c011367d5fd673332f7fa56ef81ec/.github/workflows/ci.yaml#L499-L503 in Ruff.

We should also mark these files as generated in our `.gitattributes` file

PR #415, opened 2025-05-16
[CI] Add check for changes in autogenerated files

maxmynter:
Closes: #387 
## Summary
Adds a check for changes in autogenerated files to the CI. 
This PR is taking a similar approach to [ruff](https://github.com/astral-sh/ruff/blob/cf70c7863c2c011367d5fd673332f7fa56ef81ec/.github/workflows/ci.yaml#L499-L503); executing the generation and checking for diffs. 

To have a single place where generation happens, I factored out the generation from `scripts/release.sh` to `scripts/autogenerate_files.sh` and call it in the former.

The scripts ran for me locally, but best have a close look here since I'm not too familiar with the release process. 

## Test Plan
Local testing by exectuing the `release.sh` script and using [act](https://github.com/nektos/act).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,5 @@
 * text=auto eol=lf
+
+docs/reference/cli.md linguist-generated=true
+docs/reference/configuration.md linguist-generated=true
+docs/reference/rules.md linguist-generated=true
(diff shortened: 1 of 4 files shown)


MichaReiser (approved):
Thank you

PR #432, opened 2025-05-17
Add: README badges

esadek:
Add badges to the README. Same badges as those used in the READMEs of uv and ruff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,7 @@
 # ty
 
 [![PyPI](https://img.shields.io/pypi/v/ty.svg)](https://pypi.python.org/pypi/ty)
-[![License](https://img.shields.io/pypi/l/ty.svg)](https://github.com/astral-sh/ty/blob/main/LICENSE)
 [![Supported versions](https://img.shields.io/pypi/pyversions/ty.svg)](https://pypi.python.org/pypi/ty)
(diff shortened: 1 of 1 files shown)


esadek:
SCR-20250517-miwo

zanieb:
(I kind of don't like README badges, curious what others think — I was looking at removing most of them in https://github.com/astral-sh/uv/pull/11257)

dhruvmanila:
I don't have strong opinions but if we do keep it, PyPI and Discord seems enough as I think others are present in the GitHub UI.

MichaReiser:
I'm fine with badges that serve a purpose and don't add redundant information. I agree with @dhruvmanila that I'd keep PyPI, Discord. The license is already visible on PyPI and GitHub and so is the CI status (which is rather useless on PyPI).

esadek:
Makes sense. Removed the license and CI badges.

README.md
```diff
@@ -1,5 +1,9 @@
 # ty
 
+[![PyPI](https://img.shields.io/pypi/v/ty.svg)](https://pypi.python.org/pypi/ty)
+[![Supported versions](https://img.shields.io/pypi/pyversions/ty.svg)](https://pypi.python.org/pypi/ty)
```
MichaReiser:
```suggestion
```

PR #436, opened 2025-05-17
docs: add how to start server in neovim

fabridamicelli:
Thanks for this tool!

This config worked right away for me so at least until the docs are more exhaustive, we could have this snippet.

## Summary
Add documentation for minimal setup of `ty` on Neovim.

## Test Plan
This worked out-of-the-box for me

image


hahuang65:
FWIW, I've found that even if you don't have `ty` installed (because the mason PR hasn't been accepted yet), you can still run `{ "uvx", "ty", "server" }` and get it to work, if you have `uv` installed.

In fact, I personally even run it with `poetry` if I discover a `poetry.lock` file, in order to automatically get the venv loaded up for the LSP.

```
      if require("util").dir_has_file(root_dir, "poetry.lock") then
        vim.notify_once("Running `ty` with `poetry`")
        cmd = { "poetry", "run", "uvx", "ty", "server" }
      else
        vim.notify_once("Running `ty` without a virtualenv")
        cmd = { "uvx", "ty", "server" }
      end
```

docs/reference/editor-settings.md
```diff
@@ -131,3 +131,34 @@ The extension uses the first executable that exists. This setting takes preceden
...
+
+```lua
+local configs = require('lspconfig.configs')
+configs.ty = {
+    default_config = {
+        cmd = { 'ty', 'server' },
+        filetypes = { 'python' },
+        settings = {},
+    },
+}
```
dhruvmanila: I don't think this is required given that the plugin includes the ty config: https://github.com/neovim/nvim-lspconfig/blob/ac1dfbe3b60e5e23a2cff90e3bd6a3bc88031a57/lsp/ty.lua#L4

docs/reference/editor-settings.md
```diff
@@ -131,3 +131,34 @@ The extension uses the first executable that exists. This setting takes preceden
   "ty.path": ["/home/user/.local/bin/ty"]
 }
 ```
+
+## Neovim
```
dhruvmanila: This section should actually be present under the https://github.com/astral-sh/ty/blob/main/docs/README.md#other-editors section instead. This document is only to provide reference for all editor settings.

dhruvmanila (changes requested):
I'd suggest moving the section to the relevant page and use https://docs.astral.sh/ruff/editors/setup/#neovim as a reference.

docs/reference/editor-settings.md
fabridamicelli: You are right. Now it is correct and follows instructions as in Ruff.

docs/reference/editor-settings.md
fabridamicelli: done

dhruvmanila (approved):
Thank you!

Jaehaks:
How ot configure settings in neovim? 

I wrote like this, but ty cannot recognize the rule.
or can I set config file location like ruff? ruff lsp supports `settings.configuration` to set config file globally

```lua
vim.lsp.config('ty', {
  cmd = {'ty', 'server'},
  filetypes = {'python'},
  root_dir = function (bufnr, cb)
    local root = vim.fs.root(bufnr, {
      'ty.toml',
      'pyproject.toml',
      '.git'
    }) or vim.fn.getcwd()
    -- root directory must be transferred to callback function to recognize
    cb(root)
  end,
  settings = { -- see https://docs.astral.sh/ty/reference/configuration/
    ty = {
      rules = {
        ['division-by-zero'] = 'error',
      }
    },
  },
})

```

dhruvmanila:
Currently, ty server only supports the settings mentioned in https://docs.astral.sh/ty/reference/editor-settings/ to be defined directly in an editor, we plan to add more in the future as per user feedback. I'd suggest you open a new issue to request a specific setting to be available directly in an editor if you wish so.

Regardless, you can use a configuration file (`ty.toml` or `pyproject.toml`) to specify other available settings and the server should pick that up unless it's outside the project itself.

Related: https://github.com/astral-sh/ty/issues/786

Jaehaks:
@dhruvmanila 
Thanks a lot for your reply:)

PR #440, opened 2025-05-18
Add link to playground to README.md

danielhollas:
The playground is not mentioned anywhere in this repo except in the ISSUE_TEMPLATE.

I added a link to the README in the same place and style as in Ruff repo:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,6 @@
 [![PyPI](https://img.shields.io/pypi/v/ty.svg)](https://pypi.python.org/pypi/ty)
 [![Discord](https://img.shields.io/badge/Discord-%235865F2.svg?logo=discord&logoColor=white)](https://discord.com/invite/astral-sh)
 
-[**Playground**](https://play.ty.dev/)
-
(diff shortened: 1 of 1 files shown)


MichaReiser:
Thank you. It does make sense to promote the playground but this seems a bit too prominent. I wonder if we could embed it into *getting started* instead

carljm:
I talked to someone today who was specifically looking for the playground, knew it exists, and still wasn't able to find any link to it 😆 We should definitely link it from somewhere in the README.

AlexWaygood:
I agree that the playground is a fantastic way to try out ty without having to install anything — advertising it in the README seems like a great idea to me!

danielhollas:
> knew it exists, and still wasn't able to find any link to it

Indeed that was my experience as well. 😊 

Leaving the exact placement, wording up to y'all, cheers.

carljm:
@MichaReiser How does this version look to you?

MichaReiser (approved):

PR #464, opened 2025-05-20
Add ToC and troubleshooting guidance to detailed README

brainwane:
## Summary

Add troubleshooting guidance and a table of contents to ty detailed README and installation instructions. This improves documentation navigability and, by incorporating advice from @AlexWaygood, helps prevent or remediate an error I stumbled into while testing ty at the PyCon US 2025 sprints.

## Test Plan

I previewed the Markdown file on GitHub and manually checked the formatting and hyperlinks.

PR #563, opened 2025-06-02
Update `docs/README.md`'s table of contents automatically

### LINKED ISSUE #505: TOC in README misses multiple sections
### Summary

At least, configurations, excluding files and probably more. We should double check the sections

### Version

_No response_

InSyncWithFoo:
## Summary

Resolves #505.

A new script, `update_docs_toc.py`, is added; it generates a list of links to level-2 headers and writes that back to `docs/README.md`. It is run in CI via `autogenerate_files.sh`.

## Test Plan

None.

### DIFF
--- a/scripts/update_docs_toc.py
+++ b/scripts/update_docs_toc.py
@@ -1,9 +1,13 @@
 # /// script
 # requires-python = ">=3.13"
-# dependencies = ["markdown~=3.8", "mkdocs~=1.6.1"]
+# dependencies = ["markdown", "mkdocs"]
 # ///
 
-"""Update `docs/README.md`'s TOC."""
+"""Update `docs/README.md`'s table of contents.
+
+This script generates a list of links to level-2 headers and writes that back
+to `docs/README.md`. It is run in CI via `autogenerate_files.sh`.
+"""
 
 from __future__ import annotations
 
@@ -14,15 +18,15 @@ from pathlib import Path
 from markdown import Markdown
 from mkdocs.structure.toc import AnchorLink, TableOfContents, get_toc
 
-_THIS_FILE = Path(__file__)
-_REPO_ROOT = _THIS_FILE.parent.parent
-_DOCS_README = _REPO_ROOT / "docs" / "README.md"
+THIS_FILE = Path(__file__)
+REPO_ROOT = THIS_FILE.parent.parent
+DOCS_README = REPO_ROOT / "docs" / "README.md"
 
-_TOC_START_MARKER = "<!-- TOC start -->\n\n"
-_TOC_END_MARKER = "\n\n<!-- TOC end -->"
-_TOC_SEPARATOR = " |\n"
+TOC_START_MARKER = "<!-- Generated table of contents start -->\n\n"
+TOC_END_MARKER = "\n\n<!-- Generated table of contents end -->"
+TOC_SEPARATOR = " |\n"
 
-_CODE_BLOCK = re.compile(
+CODE_BLOCK = re.compile(
     r"""(?xm)
         ^(```+).*\n
         (?>(?!\1).*\n)*?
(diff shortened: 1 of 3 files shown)


scripts/update_docs_toc.py
```diff
@@ -0,0 +1,79 @@
...
+
+from markdown import Markdown
+from mkdocs.structure.toc import AnchorLink, TableOfContents, get_toc
+
+_THIS_FILE = Path(__file__)
+_REPO_ROOT = _THIS_FILE.parent.parent
+_DOCS_README = _REPO_ROOT / "docs" / "README.md"
+
+_TOC_START_MARKER = "<!-- TOC start -->\n\n"
+_TOC_END_MARKER = "\n\n<!-- TOC end -->"
```
AlexWaygood: I think this would make it clearer to contributors that this section of the document should not be manually edited:

```suggestion
_TOC_START_MARKER = "<!-- Generated table of contents start -->\n\n"
_TOC_END_MARKER = "\n\n<!-- Generated table of contents end -->"
```

scripts/update_docs_toc.py
```diff
@@ -0,0 +1,79 @@
+# /// script
+# requires-python = ">=3.13"
+# dependencies = ["markdown~=3.8", "mkdocs~=1.6.1"]
```
AlexWaygood: instead of pinning dependencies here, maybe we could use a script lockfile, similar to [what we do for mdtest.py in the Ruff repo](https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/mdtest.py.lock) (more dogfooding!)

scripts/update_docs_toc.py
```diff
@@ -0,0 +1,79 @@
...
+_TOC_END_MARKER = "\n\n<!-- TOC end -->"
+_TOC_SEPARATOR = " |\n"
+
+_CODE_BLOCK = re.compile(
+    r"""(?xm)
+        ^(```+).*\n
+        (?>(?!\1).*\n)*?
+        \1$
+    """
+)
```
AlexWaygood: nit: I wouldn't really prefix any variable names in this file with underscores. It's a script, not a library; I don't think making a distinction between public and private names in this module is meaningful, and the underscores just add visual noise IMO

scripts/update_docs_toc.py
```diff
@@ -0,0 +1,79 @@
+# /// script
+# requires-python = ">=3.13"
+# dependencies = ["markdown~=3.8", "mkdocs~=1.6.1"]
+# ///
+
+"""Update `docs/README.md`'s TOC."""
```
AlexWaygood:
```suggestion
"""Update `docs/README.md`'s table of contents.

This script generates a list of links to level-2 headers and writes that back
to `docs/README.md`. It is run in CI via `autogenerate_files.sh`.
"""
```

AlexWaygood: It would also be great if you could add a short docstring to each function in this file. It doesn't have to be more than a sentence or two. We won't be looking at this code much, so it'll help refresh our memory about how it works when we come back to it at a later date if it's well commented.

scripts/update_docs_toc.py
```diff
@@ -0,0 +1,79 @@
...
+
+    return _TOC_SEPARATOR.join(links)
+
+
+def _replace_toc(original: str, new_toc: str) -> str:
+    start_marker_index = original.find(_TOC_START_MARKER)
+    end_marker_index = original.find(_TOC_END_MARKER)
+
+    if start_marker_index == -1 or end_marker_index == -1:
+        raise RuntimeError(f"TOC not found in {_DOCS_README}")
```
AlexWaygood:
```suggestion
    try:
        start_marker_index = original.index(_TOC_START_MARKER)
        end_marker_index = original.index(_TOC_END_MARKER)
    except ValueError as e:
        raise RuntimeError(f"No table of contents found in {_DOCS_README}") from e
```

scripts/update_docs_toc.py
```diff
@@ -0,0 +1,79 @@
...
+def _iter_links(toc: TableOfContents) -> Generator[AnchorLink]:
+    for link in toc:
+        yield link
+        yield from link.children
+
+
+def _get_new_toc(sanitized_contents: str) -> str:
+    parser = Markdown(extensions=["toc"])
+    parser.convert(sanitized_contents)
+    toc = get_toc(parser.toc_tokens)  # type: ignore  # noqa: PGH003
```
AlexWaygood, line 62: Can you add a comment explaining why you are ignoring the type-checker and linter errors here?

zanieb:
This is exactly the kind of investment I wanted to avoid; as mentioned in https://github.com/astral-sh/ty/pull/464#issuecomment-2895485130

I don't really mind that you did it now, but... we really shouldn't be spending time on things that will be obviated by moving to mkdocs in the near future.

What this indicates to me, is that we shouldn't have accepted the ToC in the first place.

AlexWaygood:
@InSyncWithFoo, thanks for the PR -- I really appreciate it! If we were planning to stick with having GitHub-hosted docs in the long term, I think we'd definitely want a solution like this.

I was already wondering whether it was worth adding this much code (which will be a maintenance burden itself!) just to keep a pretty short TOC up to date, however. Having a partially generated file in the codebase also has the disadvantage that it could confuse contributors, who have already been confused by our fully generated documentation files (which are conceptually much simpler).

Given @zanieb is also sceptical, I think we'll decline this in favour of manually updating the TOC for now. It won't be too long until we switch to mkdocs, which will mean we won't need a solution like this. And it will anyway never be possible to autogenerate all documentation to guarantee that it's kept up to date -- another recent example is https://github.com/astral-sh/ty/issues/564.

Thanks again for the PR!

PR #602, opened 2025-06-07
Fix `fstring-type-annotation` example syntax

niraj-khatiwada:
## Summary
The example syntax shown in the documentation for `fstring-type-annotation` has an extra `:` after function declaration which I think is a mistake. There should only be one `:` after return type annotation.

## Test Plan
I installed `ty` using `uv` in a fresh install and tried to go through ty rules one by one. I encountered the syntax error on `fstring-type-annotation` example code.

### DIFF
--- a/docs/reference/rules.md
+++ b/docs/reference/rules.md
@@ -233,13 +233,13 @@ Static analysis tools like ty can't analyse type annotations that use f-string n
 
 ### Examples
 ```python
-def test(): -> f"int":
+def test() -> f"int":
     ...
 ```
 
 Use instead:
 ```python
-def test(): -> "int":
+def test() -> "int":
     ...
 ```
 


AlexWaygood:
Thank you! Unfortunately these docs are generated from the doc-comments in https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/src/types/diagnostic.rs. You'll need to edit that file in the Ruff repo, run `cargo dev generate-all` locally, and then make a PR to the Ruff repo. (If you don't have `cargo` installed locally, you should also just be able to manually make the changes to https://github.com/astral-sh/ruff/blob/main/crates/ty/docs/rules.md as well as to https://github.com/astral-sh/ruff/blob/main/crates/ty_python_semantic/src/types/diagnostic.rs.) When we do our next release here, the doc updates you make to the Ruff repo will be reflected in this repo as well.

Sorry about this -- I know it's sort-of confusing!

Issue #366, opened 2025-05-13
panic: expression should belong to this TypeInference region and TypeInferenceBuilder should have inferred a type for it


jelle-openai:
### Summary

```
error[panic]: Panicked at crates/ty_python_semantic/src/types/infer.rs:714:59 when checking `<snip>.py`: `expression should belong to this TypeInference region and TypeInferenceBuilder should have inferred a type for it`
info: This indicates a bug in ty.
info: If you could open an issue at https://github.com/astral-sh/ty/issues/new?title=%5Bpanic%5D, we'd be very appreciative!
info: Platform: macos aarch64
info: Args: ["ty", "check", "."]
info: run with `RUST_BACKTRACE=1` environment variable to show the full backtrace information
info: query stacktrace:
   0: infer_definition_types(Id(8f79d5))
             at crates/ty_python_semantic/src/types/infer.rs:148
             cycle heads: infer_definition_types(Id(8f79e4)) -> 0
   1: infer_expression_types(Id(8fb907))
             at crates/ty_python_semantic/src/types/infer.rs:222
   2: infer_expression_type(Id(8fb907))
             at crates/ty_python_semantic/src/types/infer.rs:278
   3: member_lookup_with_policy_(Id(8fac2f))
             at crates/ty_python_semantic/src/types.rs:536
   4: member_lookup_with_policy_(Id(8fac2e))
             at crates/ty_python_semantic/src/types.rs:536
   5: infer_expression_types(Id(8fb973))
             at crates/ty_python_semantic/src/types/infer.rs:222
             cycle heads: member_lookup_with_policy_(Id(8fac2c)) -> 0, infer_definition_types(Id(8f79fd)) -> 0
   6: infer_expression_type(Id(8fb973))
             at crates/ty_python_semantic/src/types/infer.rs:278
   7: symbol_by_id(Id(8f9c5f))
             at crates/ty_python_semantic/src/symbol.rs:576
   8: infer_definition_types(Id(8f79e4))
             at crates/ty_python_semantic/src/types/infer.rs:148
   9: symbol_by_id(Id(8f9c5e))
             at crates/ty_python_semantic/src/symbol.rs:576
  10: class_member_with_policy_(Id(8ad30f))
             at crates/ty_python_semantic/src/types.rs:536
  11: member_lookup_with_policy_(Id(8fac2c))
             at crates/ty_python_semantic/src/types.rs:536
  12: infer_definition_types(Id(8f79fd))
[...]

PR #634, opened 2025-06-11
Bump version to 0.0.1-alpha.9

dhruvmanila:


CHANGELOG.md
```diff
@@ -1,5 +1,72 @@
...
+- Fix `--python` argument for Windows, and improve error messages for bad `--python` arguments ([#18457](https://github.com/astral-sh/ruff/pull/18457))
+
+### Bug fixes
+
+- Meta-type of type variables should be `type[..]` ([#18439](https://github.com/astral-sh/ruff/pull/18439))
+- Only consider a type `T` a subtype of a protocol `P` if all of `P`'s members are fully bound on `T` ([#18466](https://github.com/astral-sh/ruff/pull/18466))
+- Fix false positives for legacy `ParamSpec`s inside `Callable` type expressions ([#18426](https://github.com/astral-sh/ruff/pull/18426))
+- Fix panic when pulling types for `UnaryOp` expressions inside `Literal` slices ([#18536](https://github.com/astral-sh/ruff/pull/18536))
+- Fix panic when trying to pull types for attribute expressions inside `Literal` type expressions ([#18535](https://github.com/astral-sh/ruff/pull/18535))
+- Fix panic when trying to pull types for subscript expressions inside `Callable` type expressions ([#18534](https://github.com/astral-sh/ruff/pull/18534))
```
AlexWaygood, line 34: these three could possibly go under the "server" section, since I believe users would have been most likely to encounter them if they hovered over one of these AST nodes in an IDE context

CHANGELOG.md
```diff
@@ -1,5 +1,72 @@
...
+- Treat lambda functions as instances of `types.FunctionType` ([#18431](https://github.com/astral-sh/ruff/pull/18431))
+- Implement disjointness between `Callable` and `SpecialForm` ([#18503](https://github.com/astral-sh/ruff/pull/18503))
+
+### Server
+
+- Fix stale diagnostics in documents on Windows ([#18544](https://github.com/astral-sh/ruff/pull/18544))
+- Add support for `object.<CURSOR>` completions ([#18468](https://github.com/astral-sh/ruff/pull/18468))
+- Only provide declarations and bindings as completions ([#18456](https://github.com/astral-sh/ruff/pull/18456))
+- Hotfix for `expression_scope_id` panics ([#18455](https://github.com/astral-sh/ruff/pull/18455))
```
AlexWaygood: I don't think this bug ever appeared in a released version of ty, so we can probably omit the changelog entry for the bug being fixed!

```suggestion
```

AlexWaygood (approved):

CHANGELOG.md
dhruvmanila: I'm not sure if it makes sense to put them under "server" but I'm also not sure if they belong in the changelog because these were panics in our testing files. I'm fine with removing them but thought to put them under "bugs" in case anyone was facing them.

AlexWaygood: The panics were _uncovered_ by improving our test coverage, but they could easily have been triggered by a user hovering over exactly the wrong AST node in their IDE. Here's an example of exactly such a crash that was previously reported by a user: https://github.com/astral-sh/ty/issues/366#issuecomment-2909869832. It was fixed in https://github.com/astral-sh/ruff/pull/18321. So I think these definitely deserve to be listed in the changelog!

AlexWaygood: The user-facing impact right now is that users of the server are less likely to see random panics when hovering over AST nodes in their IDE, which is why I would have put them under "server". Obviously not a huge deal, though!

dhruvmanila: Oh, I see. Thanks for the context! I'll keep this in mind

PR #647, opened 2025-06-13
Document file exclusion

MichaReiser:
## Summary

Document the new `src.include` and `src.exclude` fields (not yet released)

Note: The `include` field is currently missing in the generated docs. It will exist once my overrides PR is merged (it's when I noticed that it's missing). This is hopefully before the next release.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -168,9 +168,25 @@ The Python version may also be explicitly specified using the
 
 ## Excluding files
 
-By default, ty ignores files listed in an `.ignore` or `.gitignore` file.
+ty automatically discovers all Python files in your project. You can customize where ty searches by using the [`src.include`](./reference/configuration.md#include) and [`src.exclude`](./reference/configuration.md#exclude) settings.
 
-To disable this functionality, set [`respect-ignore-files`](./reference/configuration.md#respect-ignore-files) to `false`.
+For example, with the following configuration, ty checks all Python files in the `src` and `tests` directories except those in the `src/generated` directory:
+
+```toml
+[tool.ty.src]
+include = ["src", "tests"]
+exclude = ["src/generated"]
+```
+
+By default, ty excludes a [variety of commonly ignored directories](./reference/configuration.md#exclude). If you want to include one of these directories, you can do so by adding a negative `exclude`:
+
+```toml
+[tool.ty.src]
+# Remove `build` from the excluded directories.
+exclude = ["!build"]
+```
+
+By default, ty ignores files listed in an `.ignore` or `.gitignore` file. To disable this functionality, set [`respect-ignore-files`](./reference/configuration.md#respect-ignore-files) to `false`.
 
 You may also explicitly pass the paths that ty should check, e.g.:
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -168,17 +170,54 @@ The Python version may also be explicitly specified using the
...
+By default, ty ignores files listed in an `.ignore` or `.gitignore` file. To disable this functionality, set [`respect-ignore-files`](./reference/configuration.md#respect-ignore-files) to `false`.
 
 You may also explicitly pass the paths that ty should check, e.g.:
 
 ```shell
 ty check src scripts/benchmark.py
 ```
 
-We plan on adding dedicated options for including and excluding files in future releases.
+Paths passed explicitly are checked even if they are otherwise ignored by an `exclude` or ignore file.
```
sharkdp: Maybe
```suggestion
Paths that are passed as positional arguments to `ty check` are included even if they would otherwise be ignored through `exclude` filters or an ignore-file.
```

docs/README.md
```diff
@@ -168,17 +170,54 @@ The Python version may also be explicitly specified using the
...
+
+### Include and exclude syntax
+
+Both `include` and `exclude` support gitignore like glob patterns:
+
+- `src/` matches only a directory
```
sharkdp: Should this maybe describe whether or not the directory contents would be affected?

docs/README.md
```diff
@@ -168,17 +170,54 @@ The Python version may also be explicitly specified using the
...
+- `src` matches both files and directories
+- `src` matches files or directories named `src`
```
sharkdp: This seems like a duplication?

docs/README.md
```diff
@@ -168,17 +170,54 @@ The Python version may also be explicitly specified using the
...
+- `*` matches any (possibly empty) sequence of characters (except `/`).
+- `**` matches zero or more path components.
+    This sequence **must** form a single path component, so both `**a` and `b**` are invalid and will result in an error.
+    A sequence of more than two consecutive `*` characters is also invalid.
+- `?` matches any single character except `/`
+- `[abc]` matches any character inside the brackets. Character sequences can also specify ranges of characters, as ordered by Unicode,
+    so e.g. `[0-9]` specifies any character between `0` and `9` inclusive. An unclosed bracket is invalid.
+
+Include patterns are anchored: `src` includes only `<project_root>/src` and not `<project_root>/test/src`. To include any directory named `src`, use a prefix match like so: `**/src`, but note that these can notably slow down the Python file discovery.
```
sharkdp: Maybe
```suggestion
Include patterns are anchored: The pattern `src` only includes `<project_root>/src` but not something like `<project_root>/test/src`. To include any directory named `src`, use the `**/src` prefix match. However, note that these can notably slow down the Python file discovery.
```

docs/README.md
```diff
@@ -168,17 +170,54 @@ The Python version may also be explicitly specified using the
...
+
+Unlike include patterns, exclude patterns aren't anchored unless they contain a `/`: `venv` excludes any directory named `venv`, e.g. it excludes `<project_root>/venv` and `<project_root>/sub/venv`.
```
sharkdp: > unless they contain a `/`

This is interesting. I think I would expect something like `src/generated` to apply at arbitrary depths, and only something like `/src/generated` with a leading `/` to be anchored. Maybe it's worth clarifying?

I think the examples here might not be needed?
```suggestion
Exclude patterns are *not* anchored, unless they contain a `/`: `venv` excludes all directories named `venv`, at arbitrary depths.
```

docs/README.md
```diff
@@ -168,17 +170,54 @@ The Python version may also be explicitly specified using the
...
+### Include and exclude syntax
+
+Both `include` and `exclude` support gitignore like glob patterns:
+
+- `src/` matches only a directory
+- `src` matches both files and directories
+- `src` matches files or directories named `src`
+- `*` matches any (possibly empty) sequence of characters (except `/`).
+- `**` matches zero or more path components.
+    This sequence **must** form a single path component, so both `**a` and `b**` are invalid and will result in an error.
```
dhruvmanila: I'm very confused by this examples (`**a`, `b**`), is there some kind of context that's required here? Would it be useful to include a directory tree as an example where this example is relevant?

docs/README.md
dhruvmanila: This is going to change with https://github.com/astral-sh/ruff/pull/18685, right?

MichaReiser: Yes

docs/README.md
MichaReiser: I don't think a directory tree would help here. These patterns are not valid syntax. `**` can only be used like `/**/a` but not inside a path component

Issue #86, opened 2025-03-14
[red-knot] Code completion


MichaReiser:
Implement auto-completion support

# Plan

At time of writing (2025-05-02), the plan below represents my best effort guess as to how we want to proceed here. The first few items are meant to correspond to lower hanging fruit.

One thing that is not really captured by the plan here, but I think may require some effort, is figuring out how best to identify what the current context is. I suspect there may be an easy way to start for specific cases, but the extent to which that works more generally is not totally clear to me yet.

* [ ] Add built-ins to code completions.
* [ ] Return all names from scopes in current module. [ref](https://github.com/astral-sh/ruff/pull/17744#discussion_r2069886060)
* [ ] Identify current scope and return names from just that scope.
* [ ] Identify a `Object.<typing>` context and return attributes on `Object`.
* [ ] Identify a `from module import <typing>` context and return items from `module`.

MichaReiser:
I think this is a pretty big feature and not one that specifically plays to my strengths. Why I think it's big:

* Code completion is context sensitive: It needs to list all modules when in an `import <here>` position but list a module's symbols when at the start of an identifier, or types when in a typing position (oh right, knowing whether we're in a typing position isn't trivial in Python, thanks to stringified type annotations and legacy type aliases)
* We lack the necessary queries: The module resolver allows individual module lookups but it has no API to list all modules. The same is true for all other APIs. We have APIs to lookup a single attribute but we don't have one that lists the possible attributes of a class, taking descriptor protocol, inheritance, and what not into account (and hide attributes that are hidden because of inheritance). 
* It's important to rank suggestions because you want to avoid re-experts if you can import the symbol directly.

I think we can implement something basic, but the project itself is probably a larger one that spans multiple months.

Edited:

Next steps:

* What's the minimum amount of work we can do to ship something useful?
* If we had to ship this in one week (or in two weeks), what would we do?

MichaReiser:
Completing signature might actually be a separate feature in the LSP specification and could be worth tracking separately: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_signatureHelp

MichaReiser:
@BurntSushi it might be worth updating the issue summary with the next tasks.

sharkdp:
> Offer contextual language keywords in completions

[...]

PR #650, opened 2025-06-13
docs: remove experimental section

BurntSushi:
This was only used to opt into completions. But we're removing the
opt-in in https://github.com/astral-sh/ruff/pull/18650, so let's remove
it from the docs too.

Ref #86


carljm (approved):

dhruvmanila (approved):
Thank you!

PR #673, opened 2025-06-17
Ignore default major and minor labels during version determination

zanieb:


pyproject.toml
```diff
@@ -106,6 +106,8 @@ version-files = [
 submodules = ["ruff"]
 require-labels = [{ submodule = "ruff", labels = ["ty"] }]
 ignore-labels = ["internal", "testing"]
+major-labels = []  # We do not use the major version number yet
+minor-labels = []  # We do not use the minor version number yet
```
zanieb, line 110: See defaults at https://github.com/zanieb/rooster/blob/da7f2f6e57a98f05ed2ad611dccc30c73aa1a613/src/rooster/_config.py#L29-L31

`patch-labels` is actually available there too, but unused 🙃

AlexWaygood (approved):

PR #730, opened 2025-06-30
Add documentation publishing

zanieb:

PR #729, opened 2025-06-30
Copy basic template and styling for documentation from uv

zanieb:
This copies the documentation scaffolding from uv, doing the minimum amount of work to make it viable.

PR #731, opened 2025-06-30
Publish documentation on release

zanieb:
Extends #730, separated so we can test publishing the documentation manually before doing so on release.

This is copied from ruff.


zanieb:
Needs to wait for #729

zanieb:
I changed the base to #744

PR #734, opened 2025-06-30
Build the documentation in CI

zanieb:
To ensure things are working! The `--strict` flag performs checks for valid URLs and such.


zanieb:
Our insiders keys isn't working, but it works in other repositories so I guess that's a me problem?

PR #746, opened 2025-07-01
Add favicon

zanieb:
This needs to be the Astral icon since all of the documentation pages are on the same subdomain


zanieb:
(I'll wait for #729 to land so this can land into #744)

PR #747, opened 2025-07-01
Add ty letter logo asset

zanieb:
Screenshot 2025-07-01 at 12 01 07 PM

This renders up in the top left


zanieb:
(I'll wait for #729 to land so this can land into #744)

PR #748, opened 2025-07-01
Naively split the "Overview" page into child pages

zanieb:


docs/configuration.md
```diff
@@ -0,0 +1,40 @@
...
+
+**`ty.toml`**:
+
+```toml
+[rules]
+index-out-of-bounds = "ignore"
+```
+
+> [!NOTE]
+> `ty.toml` files take precedence over `pyproject.toml` files, so if both `ty.toml` and `pyproject.toml` files are present in a directory, configuration will be read from `ty.toml`, and the `[tool.ty]` section in the accompanying `pyproject.toml` will be ignored.
```
zanieb: We'll need to rewrite these to use `!!! note` admonitions instead. This is a follow-up task noted in #744

mkdocs.template.yml
```diff
@@ -82,11 +82,21 @@ extra:
     - icon: fontawesome/brands/x-twitter
       link: https://x.com/astral_sh
 nav:
-  - Overview: index.md
+  - Introduction: index.md
+  - Concepts:
```
zanieb, line 86: We'll want an `index.md` here eventually so the header is clickable like the other ones, but we'll want to settle on the content first to avoid churn.

docs/configuration.md
sharkdp: I can take care of this tomorrow.

docs/index.md
```diff
@@ -54,408 +52,3 @@ to check the project for type errors using ty's default configuration.
...
-
-```toml
-[tool.ty.rules]
-unused-ignore-comment = "warn"
-redundant-cast = "ignore"
-possibly-unbound-attribute = "error"
-possibly-unbound-import = "error"
-```
-
-## Suppressions
```
sharkdp, line 312: This chapter was missing completely, added.

sharkdp (approved):
Thank you!

docs/index.md
zanieb: Oh tragic, it's sitting untracked in my repo — not sure why that one got missed. Thanks for catching it!

PR #779, opened 2025-07-08
Add docs for `ty.diagnosticMode`

dhruvmanila:
This PR adds documentation for the `ty.diagnosticMode` editor setting.

### DIFF
--- a/docs/reference/editor-settings.md
+++ b/docs/reference/editor-settings.md
@@ -22,6 +22,25 @@ server for features like code completion, hover, go to definition, etc.
 }
 ```
 
+## `diagnosticMode`
+
+Determines the scope of the diagnostics reported by the language server.
+
+- `workspace`: Diagnostics are reported for all files in the workspace.
+- `openFilesOnly`: Diagnostics are reported only for files that are currently open in the editor.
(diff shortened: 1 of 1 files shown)


docs/reference/editor-settings.md
```diff
@@ -22,6 +22,25 @@ server for features like code completion, hover, go to definition, etc.
 }
 ```
 
+## `diagnosticMode`
+
+Determines the scope of the diagnostics reported by the language server.
+
+- `workspace`: Diagnostics are reported for all files in the workspace.
+- `openFilesOnly`: Diagnostics are reported only for files that are currently open in the editor.
```
AlexWaygood: does it make sense to put the default first?

```suggestion
- `openFilesOnly`: Diagnostics are reported only for files that are currently open in the editor.
- `workspace`: Diagnostics are reported for all files in the workspace.
```

AlexWaygood (approved):
Thank you!

docs/reference/editor-settings.md
dhruvmanila: Either way is fine, no preference

PR #944, opened 2025-08-06
Update editor settings reference with new format

dhruvmanila:
## Summary

This PR updates the documentation to reflect the changes made in https://github.com/astral-sh/ruff/pull/19614 and https://github.com/astral-sh/ty-vscode/pull/106.

## Test Plan

<details><summary>Screenshot of the new editor settings reference page:</summary>
<p>

image

</p>
</details>

### DIFF
--- a/docs/reference/editor-settings.md
+++ b/docs/reference/editor-settings.md
@@ -4,6 +4,10 @@ The editor settings supported by ty's language server, as well as the settings s
 
 ## `python.ty.disableLanguageServices`
 
+!!! warning "Deprecated"
+
+    This option has been deprecated. Use [`ty.disableLanguageServices`](#disablelanguageservices) instead.
+
 Whether to disable the language services for the ty language server like code completion, hover,
 go to definition, etc.
 
@@ -22,6 +26,69 @@ server for features like code completion, hover, go to definition, etc.
 }
 ```
 
+______________________________________________________________________
+
+## `disableLanguageServices`
+
+Whether to disable the language services for the ty language server like code completion, hover,
+go to definition, etc.
+
+This is useful if you want to use ty exclusively for type checking and want to use another language
+server for features like code completion, hover, go to definition, etc.
+
+**Default value**: `false`
+
+**Type**: `boolean`
+
+**Example usage**:
+
+=== "VS Code"
+
+    ```json
+    {
+      "ty.disableLanguageServices": true
+    }
+    ```
+
+=== "Neovim"
+
+    ```lua
+    require('lspconfig').ty.setup({
+      settings = {
+        ty = {
+          disableLanguageServices = true,
+        },
+      },
+    })
+
+    -- For Neovim 0.11.0 and later:
+    vim.lsp.config('ty', {
+      settings = {
+        ty = {
+          disableLanguageServices = true,
+        },
+      },
+    })
+    ```
+
+=== "Zed"
+
+    ```json
+    {
+      "lsp": {
+        "ty": {
+          "settings": {
+            "ty": {
+              "disableLanguageServices": true
+            }
+          }
+        }
+      }
+    }
+    ```
+
+______________________________________________________________________
+
 ## `diagnosticMode`
 
 Determines the scope of the diagnostics reported by the language server.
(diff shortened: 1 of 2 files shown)


docs/reference/editor-settings.md
```diff
@@ -1,8 +1,79 @@
...
 
-The editor settings supported by ty's language server, as well as the settings specific to [ty's VS Code extension](https://github.com/astral-sh/ty-vscode/).
+The editor settings supported by ty's language server, as well as the settings specific to [ty's VS
+Code extension][ty-vscode].
 
-## `python.ty.disableLanguageServices`
+## Runtime settings
+
+These settings define the behavior of the language server while it is running. They can be changed
+dynamically without needing to restart the server.
```
dhruvmanila: I'm not sure if we want to provide this kind of information to the user facing docs, happy to hear others thoughts.

MichaReiser: I'd remove it. It's a nice feature, but not really relevant when configuring the editor (it only describes how ty will respond to configuration changes).

I think the bigger question here is if and how we want to group the settings.

I think I would remove *Runtime settings* but keep *Initialization options*. I'm not a huge fan of the terminology because it isn't really clear for users what initialization options are. Given that the only supported options are logging related, should we rename the section to `Logging options` (and maybe move it to the very bottom?)

docs/reference/editor-settings.md
```diff
@@ -1,8 +1,79 @@
...
+
+### `disableLanguageServices`
+
+Whether to disable the language services for the ty language server like code completion, hover,
+go to definition, etc.
```
MichaReiser: Not sure if I like it more. But this avoids using language server that many times:

```
Disables language services functionality such as code completion, hover, go to definition, etc.
```

docs/reference/editor-settings.md
```diff
@@ -1,8 +1,79 @@
...
+
+This is useful if you want to use ty exclusively for type checking and want to use another language
+server for features like code completion, hover, go to definition, etc.
```
MichaReiser:
```suggestion
This is useful if you want to use ty exclusively for type checking in combination with another language
server for features like code completion, hover, go to definition, etc.
```

docs/reference/editor-settings.md
```diff
@@ -35,13 +108,60 @@ Determines the scope of the diagnostics reported by the language server.
...
+    }
+    ```
+
+______________________________________________________________________
+
+## Initialization options
+
+The following settings are the initialization options that can be provided to the language server
+during startup. These settings are static and cannot be changed while the server is running. Any
+change to these settings requires a server restart to take effect.
```
MichaReiser: I'd rephrase or extend this section. Its current framing contains a lot of details that aren't relevant for VS code and I think also Zed users. I think it only applies to neovim users.

docs/reference/editor-settings.md
dhruvmanila: > I think I would remove _Runtime settings_ but keep _Initialization options_. I'm not a huge fan of the terminology because it isn't really clear for users what initialization options are. Given that the only supported options are logging related, should we rename the section to `Logging options` (and maybe move it to the very bottom?)

This is a good idea but I'm worried in the future we might have to rename it if there's anything other than logging related options that gets added. I think I'd keep it "initialization options" for now but I've updated the brief at the start to avoid using "language server" and change the language to be more towards using ty in an editor.

dhruvmanila: Another reason to avoid the "Logging options" section is that the `ty.trace.server` should also be part of that section but it's only specific to VS Code.

docs/reference/editor-settings.md
dhruvmanila: Yeah, good idea. I think something like "... language services that ty provides like ..." sounds reasonable (which I've updated to). What do you think?

docs/reference/editor-settings.md
dhruvmanila: They're relevant for Zed users. In general, these are relevant for all editors except for VS Code because the ty VS Code extension provides a different settings interface.

dhruvmanila:
(Going to merge this, but happy to do any follow-up if required.)

Issue #989, opened 2025-08-14
Respect parent Python interpreter that invoked ty as a candidate Python env


carljm:
This would make e.g. `uvx --with Foo ty check script_using_Foo_library.py` work as a one-liner. Today it doesn't work. uv creates an ephemeral env with Foo installed, and then invokes ty's Python shim from that env, but ty doesn't care about that. uv doesn't set `VIRTUAL_ENV`, and the ephemeral env isn't located anywhere ty's searching will find it, so ty never finds the env it was invoked from.

I suspect this would also make things "just work" for a significant number of end-user cases, where they are running ty from the same Python installation their dependencies are installed in, without `VIRTUAL_ENV` set. (Either because they have a venv and are running its binary directly without shell activation, or because they aren't using a venv at all.)

Uv has specific logic to do this (for itself), see https://github.com/astral-sh/uv/blob/a186fda2d27d74c631c50ffccb9413d31f95cc89/python/uv/__main__.py#L35 and https://github.com/astral-sh/uv/blob/2c54d3929c11fd9b5c8b0235ab201fe9d5501751/crates/uv-python/src/discovery.rs#L487-L493 -- we could borrow this approach.

We would still need to decide where in the priority this would fall. I think it would probably be last in fallback order, after explicit config and `.venv` and `VIRTUAL_ENV`.

AlexWaygood:
this used to work with previous versions of uv after a PR I made to uv, but I think they changed the way they layer ephemeral environments over the top in a recent version of uv which means that it no longer works.

MichaReiser:
CC: @zanieb

AlexWaygood:
for previous discussion, see https://github.com/astral-sh/ty/issues/320, https://github.com/astral-sh/uv/pull/13598 and https://github.com/astral-sh/ruff/pull/18335

zanieb:
That's for `uv run` not `uvx`, are you sure the behavior regressed here?

AlexWaygood:
> That's for `uv run` not `uvx`, are you sure the behavior regressed here?

Ah sorry, I'm getting mixed up between my uv executables -- yeah, I don't think `uvx` has ever worked to date; would be cool if it did!

I thought something with the `uv run` invocation broke recently but now I can't seem to find a minimal repro... I'll post a separate issue if I come across it again.

Sorry for the noise!! :-)

sharkdp:
[...]

PR #993, opened 2025-08-14
Port `__main__.py` functionality from uv

zanieb:
As part of https://github.com/astral-sh/ty/issues/989, but includes some other changes:

- Sets `VIRTUAL_ENV` if `pyvenv.cfg` can be found next to `sys.prefix`
- Sets `TY__PARENT_INTERPRETER` to `sys.executable`
- Elides tracebacks on interrupt on Windows

### DIFF
--- a/python/ty/__main__.py
+++ b/python/ty/__main__.py
@@ -6,21 +6,48 @@ import sys
 from ty import find_ty_bin
 
 
+def _detect_virtualenv() -> str:
+    """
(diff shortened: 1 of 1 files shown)


zanieb:
To be honest, I'm not sure what setting `VIRTUAL_ENV` does if we always respect `TY__PARENT_INTERPRETER` over `VIRTUAL_ENV`. That may be cruft in uv.

PR #1133, opened 2025-09-05
Update first-party environment.root docs

sharkdp:
ref. https://github.com/astral-sh/ruff/pull/20263

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -27,6 +27,10 @@ then set [`environment.root`](./reference/configuration.md#root) in your `pyproj
 root = ["./app"]
 ```
 
+Note that `python` and `tests` folders are automatically added to the project `root` if they exist,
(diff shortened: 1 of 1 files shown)


docs/modules.md
```diff
@@ -27,6 +27,10 @@ then set [`environment.root`](./reference/configuration.md#root) in your `pyproj
 root = ["./app"]
 ```
 
+Note that `python` and `tests` folders are automatically added to the project `root` if they exist,
+and if they are not packages themselves (i.e. they do not contain an `__init__.py` or a
```
AlexWaygood:
```suggestion
and if they are not packages themselves (i.e. they do not contain an `__init__.py` file or an
```

AlexWaygood (approved):
Thanks!

PR #1192, opened 2025-09-17
Upgrade to the latest rooster commit

zanieb:
Nothing pressing here, but gets us on the `main` branch and uses the same commit as uv.


zanieb:
I did yeah. It looked fine.

I just went to double check and I'm getting 502 from GitHub :(

PR #1280, opened 2025-09-29
Pin actions in release workflow to version tags

silamon:
## Summary

This workflow doesn't pin the Github Actions to a specific version using SHA tags. Now it does, just like the others, which will prevent weekly Dependabot PR's based on commit only. 

You can find the release SHA's here. 
https://github.com/actions/checkout/releases/tag/v5.0.0
https://github.com/actions/download-artifact/releases/tag/v5.0.0
https://github.com/actions/upload-artifact/releases/tag/v4.6.2

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,7 +61,7 @@ jobs:
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
-      - uses: actions/checkout@ff7abcd0c3c05ccf6adc123a8cd1fd4fb30fb493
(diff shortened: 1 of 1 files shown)


MichaReiser:
Thank you. Unfortunately, this doesn't work because cargo dist doesn't support adding the `# version` comment (CC: @Gankra). But it would certainly be nice to have.

zanieb:
Yes it does!

https://github.com/astral-sh/uv/blob/170ab1cd7fb795bdfa228162ace3328c7ccf0d12/dist-workspace.toml#L78-L82

MichaReiser:
It does support git commits (which we use in ty), but it doesn't copy over the version comment (unless the numbers of whitespaces matters)

zanieb:
Ah I see — I misread your message. Thanks for the correction :)

PR #1371, opened 2025-10-16
Bump version to 0.0.1-alpha.23

sharkdp:
I excluded the following items from the CHANGELOG. Let me know if you want one of these to be included:

- Rooster config: ignore PRs labelled `documentation` or `playground` when generating the changelog ([#1360](https://github.com/astral-sh/ty/pull/1360))
- refactor `Place` ([#20871](https://github.com/astral-sh/ruff/pull/20871))
- Standardize syntax error construction ([#20903](https://github.com/astral-sh/ruff/pull/20903))
- Use field-specifier return type as the default type for the field ([#20915](https://github.com/astral-sh/ruff/pull/20915))
- Do not bind `self` to non-positional parameters ([#20850](https://github.com/astral-sh/ruff/pull/20850))

The last one is *in principle* observable, but sounds much more scary than it is, as it only affected `CallableTypeOf[…]`, I believe.

PR #1374, opened 2025-10-16
Ignore `github-actions` bot as a contributor in CHANGELOG

sharkdp:
Re: https://github.com/astral-sh/ty/pull/1371#discussion_r2436914106

I *think* this should work?

https://github.com/zanieb/rooster/blob/b6dafd2c8dc596680bb9e91a1c52192a5ecc3a9e/src/rooster/_config.py#L43

https://github.com/zanieb/rooster/blob/b6dafd2c8dc596680bb9e91a1c52192a5ecc3a9e/src/rooster/_changelog.py#L278-L282

I don't think we need to exclude dependabot (or renovate) because those PRs are labeled as `internal` anyway.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,6 +114,7 @@ ignore-labels = [
     "documentation",
     "release",
 ]
+changelog-ignore-authors = ["github-actions"]
 major-labels = [] # We do not use the major version number yet
 minor-labels = [] # We do not use the minor version number yet
 version-format = "cargo"


AlexWaygood:
to double-check my understanding: the effect we're aiming for here is that the PRs will be listed in the changelog (because typeshed syncs are user-facing changes) but the author of the PRs (github-actions) will not be listed as one of the contributors?

sharkdp:
> to double-check my understanding: the effect we're aiming for here is that the PRs will be listed in the changelog (because typeshed syncs are user-facing changes) but the author of the PRs (github-actions) will not be listed as one of the contributors?

Correct

zanieb (approved):

PR #1636, opened 2025-11-25
pin ruff commit hashes when copying docs

oconnor663:
This is a drive-by fix/idea. I noticed that we're publishing links with line numbers that tend to go stale as the upstream Rust code is edited over time (like most of the "View source" links on [this page](https://docs.astral.sh/ty/reference/rules)). Might be nice to snapshot these when we vendor the `.md` files?

### DIFF
--- a/scripts/autogenerate_files.sh
+++ b/scripts/autogenerate_files.sh
@@ -11,14 +11,23 @@ set -eu
 script_root="$(realpath "$(dirname "$0")")"
 project_root="$(dirname "$script_root")"
 cd "$project_root"
+ruff_commit="$(git -C ruff rev-parse HEAD)"
 
 echo "Updating lockfile..."
 uv lock
 
 echo "Copying reference documentation from Ruff..."
-cp ./ruff/crates/ty/docs/cli.md ./docs/reference/
-cp ./ruff/crates/ty/docs/configuration.md ./docs/reference/
-cp ./ruff/crates/ty/docs/rules.md ./docs/reference/
-cp ./ruff/crates/ty/docs/environment.md ./docs/reference/
+copy_docs() {
+    src="$1"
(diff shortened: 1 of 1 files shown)


oconnor663:
I guess if we did want to do this, we'd need to remove the CI check that specifically tests we didn't change anything...

oconnor663:
Adding @zanieb as a reviewer. Low priority, no rush, etc. I'm curious to get your take as the CEO-at-large of Rooster :)

scripts/autogenerate_files.sh
```diff
@@ -11,14 +11,23 @@ set -eu
...
+    src="$1"
+    dest="$2"
+    [ -d "$dest" ] && dest="$dest/$(basename "$src")"
+    old_url="/github.com/astral-sh/ruff/blob/main/"
+    new_url="/github.com/astral-sh/ruff/blob/$ruff_commit/"
+    sed "s|$old_url|$new_url|g" "$src" > "$dest"
+}
+copy_docs ./ruff/crates/ty/docs/cli.md ./docs/reference/
+copy_docs ./ruff/crates/ty/docs/configuration.md ./docs/reference/
+copy_docs ./ruff/crates/ty/docs/rules.md ./docs/reference/
```
MichaReiser, line 30: I think we only need this for `rules.md`, no other markdown file contains line-references

MichaReiser (approved):

MichaReiser:
Hmm, you now run into issues with our check-generated files unedited. That script will require the same pre-processing

zanieb:
As with https://github.com/astral-sh/ruff/pull/26421 and https://github.com/astral-sh/uv/pull/16969 it seems reasonable to just not keep a checked-in copy of the generated files in which case that check can be removed.

PR #1710, opened 2025-12-01
Enable PEP 740 attestations when publishing to PyPI

woodruffw:
## Summary

This enables PEP 740 attestations on the distributions we publish to PyPI.

See https://github.com/astral-sh/uv/pull/16910 and https://github.com/astral-sh/ruff/pull/21735 for reference 🙂 

## Test Plan

Not easy to test, since it's in the release pipeline. The action itself has integration tests and has been confirmed to work on other project releases, however, so hopefully things will go smoothly here...

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -18,8 +18,7 @@ jobs:
     environment:
       name: release
     permissions:
-      # For PyPI's trusted publishing.
-      id-token: write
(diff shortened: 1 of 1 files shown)


zanieb:
Yo let's wait until it's proven working before we break all three repositories :D

woodruffw:
> Yo let's wait until it's proven working before we break all three repositories :D

Wise words, I'll mark this one as drafted until we have a successful release on uv/ruff 😅

PR #1766, opened 2025-12-05
[ty] Docs with embedded playground

sharkdp:
## Summary

No need to look. Just saving this in case we want to come back to it later.


mkdocs.template.yml
```diff
@@ -92,6 +92,9 @@ nav:
       - Rules: rules.md
       - Suppression: suppression.md
       - Editors: editors.md
+  - Features:
```
T-256, line 95: perhaps, better naming this section to `Editor features` / `LSP features` to be more explicit.

PR #1855, opened 2025-12-11
[ty] Document ty features

sharkdp:
## Summary

Document some of ty's unique type system features and add some showcases of diagnostics.

Rendered version: https://shark.fish/ty-features/features/type-checking/

### DIFF
--- a/docs/features/type-checking.md
+++ b/docs/features/type-checking.md
@@ -82,7 +82,13 @@ def greet(being: Person | Animal | None):
         print("Hello there!")
 ```
 
-(Full example in the [playground](https://play.ty.dev/117008af-e21b-4d25-ab4c-c6dcbbfa1310))
+(Full example in the [playground](https://play.ty.dev/31f2c718-516a-4a85-80e0-2a4682b818f1))
+
+!!! info
+
+    If you run into a situation like this and would like `Animal` to be excluded from the narrowed
+    type as well, you can make `Animal` a `@final` class. This also allows ty to infer a more precise
+    type for `being.name` (`str` instead of `object`).
 
 If you only use ty as a type checker, you can also make direct use of intersection types in
 annotations by importing `Intersection` from the special `ty_extensions` module that is (currently)


docs/features/language-server.md
```diff
@@ -0,0 +1,19 @@
...
+- Find references
+- Inlay hints
+- TODO: Baking of inlay hints into the source code
+- Signature help
+- Highlight references of symbols
+- Semantic highlighting
+
+And, the following experimental features:
+
+- Symbol renaming (`ty.experimental.rename`)
```
sharkdp: No, this is not in its final form.

docs/features/type-checking.md
```diff
@@ -0,0 +1,223 @@
...
+
+        return str({
+            "data": obj.serialize_json(),
+            "version": obj.version
+        })
+    else:
+        return obj.serialize_json()
+```
+
+(Full example in the [playground](https://play.ty.dev/39241435-5e78-4ce9-817f-ce65be73a6ed))
```
carljm, line 40: I'm not sure if this implies any change to this section, but it's worth noting that all major type checkers have enough "implicit synthesized intersection" support that they support this example as well as we do. So this isn't a unique ty feature.

docs/features/type-checking.md
```diff
@@ -0,0 +1,223 @@
...
+
+Intersections can also be built using gradual types like `Any` or its implicit counterpart
+`Unknown`. For example, imagine you call into untyped (third party) code that returns an object of
+type `Unknown`. Narrowing the type of that object using `isinstance` will result in an intersection
+type `Unknown & Iterable`. This type allows you to use `obj` as an iterable, while still giving you
+access to attributes defined on the original unknown type:
```
carljm: I find this example slightly odd, because iterating over `Unknown` is also fine. It returns `Unknown`, whereas with the narrowing we iterate over `Iterable[object] & Unknown`, which returns `object & Unknown` (or it should -- it actually today returns a Todo type from calling an intersection) -- which simplifies to just `Unknown`. The intersection here would be more meaningful if it allowed us to have useful type information about some attributes that otherwise we wouldn't have?

I guess maybe I'm focusing on the wrong part of this sentence. The part that's unique to ty isn't "allows you to use `obj` as an iterable" (all type checkers do that), it's letting you use other attributes of the unknown type (no other type checker does that.)

docs/features/type-checking.md
```diff
@@ -0,0 +1,223 @@
...
+    species: str
+
+def greet(being: Person | Animal | None):
+    if hasattr(being, "name"):
+        # `being` is now of type `Person | (Animal & <Protocol with members 'name'>)`
+
+        print(f"Hello, {being.name}!")
+    else:
+        print("Hello there!")
+```
```
carljm, line 83: This is such a great example! No other type checker can do this fully. Pyrefly gets the closest (it's the only one that allows using the `name` attribute), but it still thinks `None` is a possible type. Might be interesting to extend the example such that excluding `None` becomes relevant? (It's not a type error to string-format `None`).

docs/features/type-checking.md
```diff
@@ -0,0 +1,223 @@
...
+
+(Full example in the [playground](https://play.ty.dev/53173746-9676-4b63-b87a-88892f10a2a9))
```
carljm: I happened to notice that this playground link includes the `untyped_module.py` from the previous example still

docs/features/type-checking.md
```diff
@@ -0,0 +1,223 @@
...
+        # reveals: (Item & Top[list[Unknown]]) | list[Item]
+        reveal_type(items)
+```
+
+(Full example in the [playground](https://play.ty.dev/f6f28717-1020-4bc6-9365-2717af89b4c3))
+
+You might think that the type inside the `if` branch should just be `list[Item]`. What's happening
+instead is that ty tells us that `items` could also be a subclass of `Item` *and* `list`!
+
+While this is correct, you might still want to distinghuish between "actual" `Item`s and lists of
```
carljm:
```suggestion
While this is correct, you might still want to distinguish between "actual" `Item`s and lists of
```

docs/features/type-checking.md
```diff
@@ -0,0 +1,223 @@
...
+(Full example in the [playground](https://play.ty.dev/a5286db1-cdfd-45e7-af54-29649ba5c423))
+
+This is achieved by treating `max_retries` as being of type `Unknown | None`, which means that the
+type of the attribute is not fully known, but `None` is definitely a possible value.
+
+Users can always opt into stricter checking by adding type annotations (`int | None`, in this case).
+
+!!! info
+
+    We are also planning to add a mode for users that prefer to have stricter types in these
```
carljm:
```suggestion
    We are also planning to add a mode for users that prefer to have stricter types inferred by default in these
```

docs/features/type-checking.md
```diff
@@ -0,0 +1,223 @@
...
+as unreachable:
+
+```py
+if 3**2 + 4**2 == 5**2:
+    answer = "equation holds"
+else:
+    answer = "equation does not hold"  # unreachable
+
+reveal_type(answer)  # reveals: Literal["equation holds"]
+```
```
zsol: I liked that every single example & paragraph so far was genuinely relatable - it made reading this top to bottom flow really well. To me, this example didn't really add anything, and interrupted this flow. Maybe consider removing it and going straight to the pydantic example below?

docs/features/type-checking.md
```diff
@@ -0,0 +1,223 @@
...
+Users can always opt into stricter checking by adding type annotations (`int | None`, in this case).
+
+!!! info
+
+    We are also planning to add a mode for users that prefer to have stricter types in these
+    situations. You can follow [this issue](https://github.com/astral-sh/ty/issues/1240) for updates.
+
+## Fixpoint iteration
+
+ty uses fixpoint iteration when inferring types for self-referential code. For example, the type of
```
zsol: The term "self-referential code" (and maybe even fixpoint iteration) could use some elaboration here IMO
Actually I wonder if we want users to learn the term "fixpoint iteration" - is the takeaway here mainly that ty can handle code that references itself?

docs/features/type-checking.md
carljm: Not relatable?? You don't use your type-checker to [brute-force Diophantine equations](https://play.ty.dev/00ac7e54-1e32-47b1-b18e-a786a6cee0b8)?

docs/features/type-checking.md
```diff
@@ -0,0 +1,223 @@
...
+
+(Full example in the [playground](https://play.ty.dev/f6f28717-1020-4bc6-9365-2717af89b4c3))
+
+You might think that the type inside the `if` branch should just be `list[Item]`. What's happening
+instead is that ty tells us that `items` could also be a subclass of `Item` *and* `list`!
+
+While this is correct, you might still want to distinghuish between "actual" `Item`s and lists of
+`Item`s. You can achieve this either by making `Item` a `@final` class, or by doing the `isinstance`
+check against `Item` instead. In the latter case, the `else` branch will have a narrowed type of
+`list[Item] & ~Item`, which acts effectively like `list[Item]`.
```
zsol: nit: consider changing the example to make `Item` `@final`, and adding an info box at the bottom about why it's there - it was a bit puzzling to read the code snippet, trying to work out what the type expression in the comment means before reading the answer below

docs/features/type-checking.md
zsol: Only when my coworkers force me to!

AlexWaygood: Watch out @carljm, he'll call HR on us again

docs/features/type-checking.md
sharkdp: That's a great suggestion. Thank you.

docs/features/type-checking.md
sharkdp: > I guess maybe I'm focusing on the wrong part of this sentence. The part that's unique to ty isn't "allows you to use `obj` as an iterable" (all type checkers do that), it's letting you use other attributes of the unknown type (no other type checker does that.)

Yes, exactly. I tried to rephrase it to make that clearer.

docs/features/type-checking.md
sharkdp: Yes, I know :neutral_face:. I still found it useful to start with this basic example? The examples below demonstrate things that other type checkers can't do, but they are quite a bit more involved.

docs/features/type-checking.md
sharkdp: > Might be interesting to extend the example such that excluding `None` becomes relevant? (It's not a type error to string-format `None`).

I don't really understand what pyrefly is doing there, to be honest. It thinks that `being` is still of type `None`, but it still allows you to access `being.name`. But that attribute has a type of `Unknown`, so I guess there's no way to "force" pyrefly to fail here. If you look closely, it reveals a type of `Animal | Person | None (_.name: Unknown)`, so that `(_.name: Unknown)` part might be some sort of intersection-like behavior?

docs/features/type-checking.md
AlexWaygood: I might be tempted just to omit this section entirely... The fact that we use fixpoint iteration is interesting to other type-checker authors and compiler devs, but I'm not sure it will be that interesting to users. The _results_ of it _might_ be interesting to users -- it's kinda cool that we can infer `Literal[0, 1, 2, 3, 4] | Unknown` for `LoopingCounter.value`. But is that _significantly_ better than [mypy's inferred type](https://mypy-play.net/?mypy=latest&python=3.12&gist=5a85116acffab621fdc2df96ca6d5df9) of `int`, such that it's worth highlighting for users?

sharkdp: Talked about the same topic with Carl yesterday. We agree that it's not very useful, but it's just extremely cool, and I don't think there's any harm in leaving this at the bottom of this document here. I rephrased this section to hopefully make it clearer.

docs/features/type-checking.md
```diff
@@ -0,0 +1,216 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from ty_extensions import Intersection
+
+    type SerializableVersioned = Intersection[Serializable, Versioned]
+
+def output_as_json(obj: SerializableVersioned) -> str:
+    ...
+```
```
AlexWaygood, line 107: there _are_ ways of using this even if you support multiple type checkers, though it obviously gets more fiddly. For example, this snippet passes mypy out-of-the-box, because mypy treats any `MYPY` variables as always-true, the same as it does for `TYPE_CHECKING`. It doesn't pass pyright out of the box, but I believe it _would_ pass pyright if you had a pyright configuration file with `{"PYRIGHT": true}` specified for the `defineConstant` setting (https://microsoft.github.io/pyright/#/configuration?id=environment-options).

```py
from typing import TYPE_CHECKING

MYPY = False

class Serializable: ...
class Versioned: ...

if TYPE_CHECKING:
    if MYPY or PYRIGHT:
        type SerializableVersioned = Serializable
    else:
        from ty_extensions import Intersection
    
        type SerializableVersioned = Intersection[Serializable, Versioned]

def output_as_json(obj: SerializableVersioned) -> str:
    return "foo"
```

not sure whether or not it's worth mentioning this; it might be too complicated for it to be worth it

sharkdp: Right. I guess we might include something like this if we have our own way of specifying a "this is being type checked by ty" test, but excluding every other type checker that you're using feels like a stretch.

AlexWaygood: Yeah. I feel quite sure we have an issue open _somewhere_ to add a ty equivalent of the mypy/pyright setting that enables this feature (the mypy config option is https://mypy.readthedocs.io/en/stable/config_file.html#confval-always_true). But who knows where that issue is in our list of 543...

docs/features/type-checking.md
carljm: Pyrefly is just applying its equivalent of our "place" narrowing. So it still thinks the type of `being` is `Animal | Person | None`, but it additionally knows that `being.name` exists with type `Unknown`.

carljm: > I guess there's no way to "force" pyrefly to fail here

What I was thinking of was not something based on the type of `being.name`, but something based on us knowing that `None` is excluded. For example, if `Person` and `Animal` both had a `greet` method, which we call as part of printing the greeting. This call would fail on `None` but succeed on `Person` and `Animal`.

docs/features/type-checking.md
carljm: Yes, I think that makes sense.

docs/features/type-checking.md
sharkdp: > What I was thinking of was not something based on the type of `being.name`, but something based on us knowing that `None` is excluded.

:+1: 

> For example, if `Person` and `Animal` both had a `greet` method, which we call as part of printing the greeting. This call would fail on `None` but succeed on `Person` and `Animal`.

Hm, I think I would find that a little contrived? Why would you check `hasattr(…, "name")` only to call a `.greet` method on the object? Or would we pass in the object's name into its own `.greet` method? That seems weird on its own... and I would still find it strange that the code relies on the implicit knowledge that every class in that union that has a `name` attribute also has a `greet` method. Happy to change it though if we can come up with something that looks a little bit more like it could be realistic code.

What's also related to the exclusion of `None` is the fact that we can infer the type of `being.name` as `str` as soon as you make `Animal` final. All other type checkers either infer `Any` or a union `Unknown | str`. Maybe that's something we could mention?

sharkdp: I added a new "Info: …" box that explains this, and changed the playground example to include a comment that encourages readers to play with adding `@final`.

docs/features/type-checking.md
```diff
@@ -0,0 +1,222 @@
...
+def to_json(person: Person):
+    if PYDANTIC_V2:
+        return person.model_dump_json()  # no error here when checking with 1.x
+    else:
+        return person.json()
+```
+
+(Full example in the [playground](https://play.ty.dev/34a227bb-93d5-405e-86c3-72f57ec5642e))
+
+## Gradual guarantee
```
danielhollas, line 171: Should Gradual guarantee be at the top? Compared to the other somewhat more theoretical concepts, this one seems like it has very clear implications on how ty behaves in general.

(Sorry for a drive by comment, this document is super cool. Looking forward to the beta release! 🚀)

sharkdp: Thank you for your comment!

It's somewhat deliberately *not* at the top, because we're not 100% sure yet if our current behavior will be the same once we release a stable version. If the answer is yes, we can/should certainly move this section to the top.

sharkdp:
@zanieb I merged this already, but feel free to add more review comments, and I will address them in a follow-up. I wanted to publish the latest version, but I don't have the necessary permissions to merge https://github.com/astral-sh/docs/pull/308. See these pages for the latest rendered version:

* https://ca6eabdc.docs-119.pages.dev/ty/features/type-checking/
* https://ca6eabdc.docs-119.pages.dev/ty/features/diagnostics/

docs/features/type-checking.md
```diff
@@ -0,0 +1,222 @@
...
+
+(Full example in the [playground](https://play.ty.dev/31f2c718-516a-4a85-80e0-2a4682b818f1))
+
+!!! info
+
+    If you run into a situation like this and would like `Animal` to be excluded from the narrowed
+    type as well, you can make `Animal` a `@final` class. This also allows ty to infer a more precise
+    type for `being.name` (`str` instead of `object`).
+
+If you only use ty as a type checker, you can also make direct use of intersection types in
```
zsol, line 93: nit: `If ty is the only type checker you use` (you can use ty as a type checker as well as a code navigation tool)

sharkdp: Thank you => https://github.com/astral-sh/ty/pull/1890

PR #1869, opened 2025-12-12
update implicit root docs

Gankra:
As of https://github.com/astral-sh/ruff/pull/21817, ./tests are no longer implicitly included.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -27,7 +27,7 @@ then set [`environment.root`](./reference/configuration.md#root) in your `pyproj
 root = ["./app"]
 ```
 
-Note a `./python` folder is automatically added to the project `root` if it exists,
+Note that a `./python` folder is automatically added to the project `root` if it exists,
 and is not itself a package (i.e. does not contain an `__init__.py` file or an
 `__init__.pyi` file).
 


carljm:
Oh sorry, didn't realize this also changes a non-generated file.

docs/modules.md
```diff
@@ -27,8 +27,8 @@ then set [`environment.root`](./reference/configuration.md#root) in your `pyproj
 root = ["./app"]
 ```
 
-Note that `python` and `tests` folders are automatically added to the project `root` if they exist,
-and if they are not packages themselves (i.e. they do not contain an `__init__.py` file or an
+Note that a `./python` folder is automatically added to the project `root` if it exists,
+and is not itself a package (i.e. does not contain an `__init__.py` file or an
```
carljm, line 31: Why do we discuss this down here, separately from the mention of `./src` above? Isn't the treatment of `./python` the same as the treatment of `./src`? (I realize that choice predates this PR.)

docs/modules.md
carljm, line 8:
```suggestion
By default, ty searches for first-party modules in the project's root directory and in `src` and `python`
subdirectories, if the latter are present and are not packages (don't contain an `__init__.py` or `__init__.pyi` file).
```

docs/modules.md
Gankra: Hmm other variants of this doc seemed to suggest there were two mutually exclusive "layout modes" we detected, but the ./python addition was applicable to both.

carljm: I don't think that is the case...

Gankra:
Oh ha, I had auto-merge set from before Carl had comments, and Alex dismissing his review unblocked it, whoops!

Issue #1825, opened 2025-12-09
Correctly handle variadic parameters for overload step 5 filtering


dhruvmanila:
### Summary

https://github.com/astral-sh/ruff/pull/21859 doesn't correctly fix the underlying issue and there are multiple TODOs in the test: https://github.com/astral-sh/ruff/blob/b21d520f067f1add859bf6494e28a3c5c7a020f4/crates/ty_python_semantic/resources/mdtest/call/overloads.md#varidic-argument-with-generics that needs to be resolved.

dhruvmanila:
(I'm moving this into Stable as resolving the cycle during function inference is more important to resolve.)

fosskers:
A sample reproduction of what currently doesn't work:
```python
class Foo[T]:
    pass


class Bar[*U]:
    pass


Baz = Foo[Bar[int]]  # <- ty claims that `Bar` can't be subscripted
```

AlexWaygood:
@fosskers that's unrelated to this issue — that's https://github.com/astral-sh/ty/issues/156

Issue #1741, opened 2025-12-03
Emit diagnostics for unsupported comparisons involving non-fixed-length tuples


AlexWaygood:
Pyright, mypy and pyrefly all emit diagnostics on all four comparisons here, but ty currently does not emit any diagnostics on this snippet:

```py
def f(
    a: tuple[int, ...],
    b: tuple[str, ...],
    c: tuple[str]
):
    a < b
    b < a
    a < c
    c < a
```

AlexWaygood:
we already emit diagnostics if both operands are _fixed_-length tuples, e.g.

```py
def f(
    x: tuple[int, int],
    y: tuple[str]
):
    x < y  # error: [unsupported-operator]
```

Issue #1675, opened 2025-11-29
No diagnostic reported for bad use of `@override` if the method has other decorators too


AlexWaygood:
Similar to #1674. We do not currently issue a complaint about this snippet, but pyre (the predecessor to pyrefly, and the reference implementation for PEP 698) [does](https://pyre-check.org/play?input=from%20typing%20import%20override%0Afrom%20functools%20import%20lru_cache%0A%0Aclass%20Bar%3A%0A%20%20%20%20%40override%0A%20%20%20%20%40lru_cache%0A%20%20%20%20def%20method(self%2C%20x%3A%20int)%20-%3E%20None%3A%20...):

```py
from typing import override
from functools import lru_cache

class Bar:
    @override
    @lru_cache
    def method(self, x: int) -> None: ...
```

The reason why we issue no complaint is that the `lru_cache` decorator transforms the function definition into an instance of `_lru_cache_wrapper`. We currently only recognise methods as being `@override` if they have `Type::FunctionLiteral` types in our internal model, so the transformation into an instance of `_lru_cache_wrapper` means that we "forget" that the method was decorated with `@override`. To fix this issue, we may need to treat the `@override` decorator more like a type qualifier that "survives" the transformations from a function-literal type to a callable type and then to an `_lru_cache_wrapper` type.

[...]

Issue #1535, opened 2025-11-12
Support Concatenate special form


carljm:
https://typing.python.org/en/latest/spec/generics.html#paramspec

vlashada:
`ty` fails to infer the return type (`T`) for a generic method whose first parameter is a callable typed with `Callable[Concatenate[SelfType, P], T]`. In practice this shows up in `polars.DataFrame.pipe`, where `reveal_type(df.pipe(f))` becomes `Unknown` even when `f` is fully annotated to return `DataFrame`.

Polars documents `DataFrame.pipe` as:

```py
DataFrame.pipe(
    function: Callable[Concatenate[DataFrame, P], T],
    *args: P.args,
    **kwargs: P.kwargs,
) -> T
```

## Reproduction (no third-party deps)

https://play.ty.dev/60c9407f-0310-4b4e-b078-fba3d381f132

```py
from typing import Callable, Concatenate, ParamSpec, TypeVar, reveal_type

P = ParamSpec("P")
T = TypeVar("T")

class DF:
    def pipe(self, function: Callable[Concatenate["DF", P], T], *args: P.args, **kwargs: P.kwargs) -> T:
        return function(self, *args, **kwargs)

def f(df: DF) -> DF:
    return df

reveal_type(DF())          # expected: DF
reveal_type(DF().pipe(f))  # expected: DF; actual (ty): Unknown
````

## Reproduction (real-world: Polars)

```py
import polars as pl
from typing import reveal_type

def f(df: pl.DataFrame) -> pl.DataFrame:
    return df

reveal_type(pl.DataFrame())          # expected: DataFrame
reveal_type(pl.DataFrame().pipe(f))  # expected: DataFrame; actual (ty): Unknown
```

dhruvmanila:
@vlashada Thanks for the report. The issue that you're highlighting is actually https://github.com/astral-sh/ruff/pull/21551. I tried out your example (the "no third-party deps" one) and it works:

```
info[revealed-type]: Revealed type
  --> /Users/dhruv/playground/ty/bug-report/main.py:21:13
   |
21 | reveal_type(DF())  # expected: DF
   |             ^^^^ `DF`
22 | reveal_type(DF().pipe(f))  # expected: DF; actual (ty): Unknown
   |

info[revealed-type]: Revealed type
  --> /Users/dhruv/playground/ty/bug-report/main.py:22:13
   |
21 | reveal_type(DF())  # expected: DF
[...]

PR #1876, opened 2025-12-13
Type system reference

sharkdp:
A document in this style is probably waaay too detailed for users, but it's sort of helpful for us?

This document draws from various sources:
* https://typing.python.org/en/latest/
* https://peps.python.org/topic/typing/, and all included PEPs
* Our mdtests
* Our actual Rust source code
* ty issue tracker

### DIFF
--- a/docs/reference/type-system-features.md
+++ b/docs/reference/type-system-features.md
@@ -1,6 +1,6 @@
 # Type system features
 
-This issue summarizes the support for various type system features in ty. Sections are organized to follow the structure of the [Python typing specification](https://typing.python.org/en/latest/spec/), with some additional sections at the end.
+This issue summarizes the support for various type system features in ty. Sections are organized to follow the structure of the [Python typing specification](https://typing.python.org/en/latest/spec/), with some additional sections at the end. If a top-level item is marked completed without any sub-items, you can generally expect that feature to be fully implemented (and we value any bug reports in case you find issues). If a top-level item is marked completed with open sub-items, the feature is generally working, but there might be some open issues — including but not limited to the ones listed (in this case, we also value any bug reports, but please search the issue tracker before doing so). If a top-level item is not checked, the feature is not implemented yet (it's probably not helpful to report bugs, but feel free to upvote the tracking issue or comment if you have useful information).
 
 ## Special types and type qualifiers
 
@@ -54,10 +54,13 @@ This issue summarizes the support for various type system features in ty. Sectio
 - [x] `ParamSpec` defaults
 - [x] `Self`
     - [ ] `Self` in attribute annotations #1124
-- [ ] `TypeVarTuple` #156
-- [ ] `Callable`s in typevar solving
-- [ ] Generic protocols in typevar solving #1714
-- [ ] Generic bounds/constraints on type variables #1839
+- [ ] Solve type variables in all cases #623
+    - [x] Generic classes
+    - [x] Unions
+    - [ ] `Callable`s
+    - [ ] Generic protocols #1714
+    - [ ] Generic bounds/constraints on type variables #1839
+- [ ] Support `TypeVarTuple` #156
 
 ## Protocols
 


docs/reference/type-system-features.md
```diff
@@ -0,0 +1,258 @@
...
+| Protocol inheritance                          | ✅                                                              |
+| `is_protocol()`, `get_protocol_members()`     | ✅                                                              |
+| `@runtime_checkable` decorator                | ✅                                                              |
+| Protocol instantiation restriction            | ✅                                                              |
+| Non-protocol class inheritance restriction    | ✅                                                              |
+| `@property` members                           | ⚠️ partial [#1379](https://github.com/astral-sh/ty/issues/1379) |
+| `@classmethod` and `@staticmethod` members    | ❌ [#1381](https://github.com/astral-sh/ty/issues/1381)         |
+| `ClassVar` members                            | ❌ [#1380](https://github.com/astral-sh/ty/issues/1380)         |
+| `type[SomeProtocol]`                          | ❌ [#903](https://github.com/astral-sh/ty/issues/903)           |
+| Modules as protocol implementations           | ❌ [#931](https://github.com/astral-sh/ty/issues/931)           |
```
AlexWaygood:
```suggestion
| Modules as protocol implementations           | ⚠️ [#931](https://github.com/astral-sh/ty/issues/931)           |
```

It's only if you use a method member that things don't work as expected. And arguably that's unsound anyway (i have some issues with the spec here 😆). But we may need to find a solution for compatibility sake anyway.

docs/reference/type-system-features.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Overloads
```
AlexWaygood, line 242: Worth mentioning https://github.com/astral-sh/ty/issues/1825?

docs/reference/type-system-features.md
```diff
@@ -0,0 +1,258 @@
...
+| `is None` / `is not None` narrowing       | ✅                                                      |
+| `is` / `is not` identity narrowing        | ✅                                                      |
+| Truthiness narrowing                      | ✅                                                      |
+| `assert` narrowing                        | ✅                                                      |
+| `match` statement narrowing               | ✅                                                      |
+| `hasattr()` narrowing                     | ✅                                                      |
+| `callable()` narrowing                    | ✅                                                      |
+| Assignment narrowing                      | ✅                                                      |
+| `TypeIs[…]` user-defined type guards      | ✅                                                      |
+| `TypeGuard[…]` user-defined type guards   | ⚠️ partial (no narrowing applied yet)                   |
```
AlexWaygood: I think from a user perspective this is basically the same as "not supported"?

```suggestion
| `TypeGuard[…]` user-defined type guards   | ❌ (no narrowing applied yet)                   |
```

docs/reference/type-system-features.md
```diff
@@ -0,0 +1,258 @@
...
+| `None`                                            | ✅     |
+| `NoReturn`, `Never`                               | ✅     |
+| `object`                                          | ✅     |
+| `Literal[...]` (strings, ints, bools, enum, None) | ✅     |
+| `LiteralString`                                   | ✅     |
+| `Callable[[...], R]`                              | ✅     |
+| `Callable[..., R]` (arbitrary arguments)          | ✅     |
+| `Callable` with `ParamSpec`                       | ✅     |
+| `type[C]`                                         | ✅     |
+| `Final`, `Final[T]`                               | ✅     |
```
AlexWaygood: This should be "partial", we still don't emit an error if you override a `Final` attribute on a subclass

docs/reference/type-system-features.md
```diff
@@ -0,0 +1,258 @@
...
+| `@final` decorator                                | ✅     |
+| `ClassVar`, `ClassVar[T]`                         | ✅     |
+| `InitVar[T]` (see Dataclasses)                    | ✅     |
+| `Annotated[T, ...]`                               | ✅     |
+| `Required[T]`, `NotRequired[T]` (see TypedDict)   | ✅     |
+| `ReadOnly[T]` (see TypedDict)                     | ✅     |
+| `Union[X, Y]`, \`X                                | Y\`    |
```
AlexWaygood: Looks like the formatting went awry here

docs/reference/type-system-features.md
```diff
@@ -0,0 +1,258 @@
...
+| `Optional[X]`                                     | ✅     |
+
+## Type aliases
+
+[Official documentation](https://typing.python.org/en/latest/spec/aliases.html)
+
+| Feature                                                 | Status                                                              |
+| ------------------------------------------------------- | ------------------------------------------------------------------- |
+| Implicit type aliases (`Alias = int`)                   | ✅                                                                  |
+| PEP 613 `TypeAlias` annotation                          | ✅                                                                  |
```
AlexWaygood: Should be partial, we still don't support fully stringified r.h.s. of a PEP-613 alias

docs/reference/type-system-features.md
```diff
@@ -0,0 +1,258 @@
...
+| --------------------------- | ------ |
+| `cast(T, value)`            | ✅     |
+| `assert_type(value, T)`     | ✅     |
+| `assert_never(value)`       | ✅     |
+| `reveal_type(value)`        | ✅     |
+| `TYPE_CHECKING` constant    | ✅     |
+| `no_type_check` decorator   | ✅     |
+| `type: ignore` comments     | ✅     |
+| `@deprecated` decorator     | ✅     |
+| `@override` decorator       | ✅     |
```
AlexWaygood: Only partial, we still don't support the stricter version of the check that is also specified

docs/reference/type-system-features.md
```diff
@@ -0,0 +1,258 @@
...
+| Redundant `cast` diagnostic | ✅     |
+
+## Invalid overrides
+
+(Liskov Substitution Principle)
```
AlexWaygood: Things we currently check:
- method overridden by method

Things we don't yet check:

- Method overridden by non-method
- Non-method overridden by non-method

docs/reference/type-system-features.md
```diff
@@ -0,0 +1,258 @@
...
+| `TypeVar` defaults (PEP 696)                      | ✅                                                                       |
+| `TypeVar` variance (`covariant`, `contravariant`) | ✅                                                                       |
+| `TypeVar` variance inference (`infer_variance`)   | ✅                                                                       |
+| Generic classes (legacy and PEP 695 syntax)       | ✅                                                                       |
+| Generic functions (legacy and PEP 695 syntax)     | ✅                                                                       |
+| Generic type aliases (PEP 695)                    | ⚠️ some limitations [#1851](https://github.com/astral-sh/ty/issues/1851) |
+| `ParamSpec` (legacy and PEP 695 syntax)           | ✅                                                                       |
+| `ParamSpec.args`, `ParamSpec.kwargs`              | ✅                                                                       |
+| `Concatenate`                                     | ✅                                                                       |
```
dhruvmanila: `Concatenate` is not supported yet https://github.com/astral-sh/ty/issues/1535

docs/reference/type-system-features.md
sharkdp: Do we have a ticket for this? What about https://github.com/astral-sh/ty/issues/1675?

AlexWaygood: Just the original ticket #155, which was never closed.

> What about https://github.com/astral-sh/ty/issues/1675?

that one's not important enough to be worth mentioning here IMO, but no strong opinion!

docs/reference/type-system-features.md
```diff
@@ -2,41 +2,123 @@
...
+| `TypeVar` variance inference (`infer_variance`)   | ✅                                                                       |
+| Generic classes (legacy and PEP 695 syntax)       | ✅                                                                       |
+| Generic functions (legacy and PEP 695 syntax)     | ✅                                                                       |
+| Generic type aliases (PEP 695)                    | ⚠️ some limitations [#1851](https://github.com/astral-sh/ty/issues/1851) |
+| `ParamSpec` (legacy and PEP 695 syntax)           | ✅                                                                       |
+| `ParamSpec.args`, `ParamSpec.kwargs`              | ✅                                                                       |
+| `ParamSpec` defaults                              | ✅                                                                       |
+| `TypeVarTuple`                                    | ❌ [#156](https://github.com/astral-sh/ty/issues/156)                    |
+| `Self`                                            | ✅                                                                       |
```
AlexWaygood: I think we can only claim partial support while https://github.com/astral-sh/ty/issues/1124 is open

docs/reference/type-system-features.md
```diff
@@ -143,72 +240,64 @@ This page summarizes the support for various type system features in ty.
...
+| Inheritance, generic dataclasses, descriptor-typed fields                                                                         | ✅                                                      |
+| `replace()`, `__replace__`                                                                                                        | ⚠️ `__replace__` works, `replace()` returns `Unknown`   |
+| `asdict()`                                                                                                                        | ⚠️ incorrectly accepts class objects                    |
+| `astuple()`                                                                                                                       | ⚠️ not tested                                           |
+| `make_dataclass()`, `is_dataclass()`                                                                                              | ⚠️ not tested                                           |
+| Diagnostic: frozen dataclass inheriting from non-frozen                                                                           | ⚠️ not tested                                           |
```
AlexWaygood: This was added over the weekend in https://github.com/astral-sh/ruff/commit/a544c59186731d8cab4c5cce05a228af585fbafe!

I think we can check it off now

docs/reference/type-system-features.md
```diff
@@ -228,18 +317,65 @@ This page summarizes the support for various type system features in ty.
...
+| `__all__` declarations                            | ✅                                                      |
+| `__all__` mutations (`.append`, `.extend`, `+=`)  | ✅                                                      |
+| `__all__` with submodule `__all__`                | ✅                                                      |
+| Wildcard (`*`) imports                            | ✅                                                      |
+| Wildcard imports in stubs re-export               | ✅                                                      |
+| Re-export conventions (`import X as X`)           | ✅                                                      |
+| Conditional re-exports in stub files              | ✅                                                      |
+| Reachability constraints in imports               | ✅                                                      |
+| Cyclic imports                                    | ✅                                                      |
+| `py.typed` marker files                           | ⚠️ ignored for type checking                            |
```
AlexWaygood: I think we respect these as much as we ever intend to? I'd be inclined to check this one off

docs/reference/type-system-features.md
```diff
@@ -256,3 +392,34 @@ This page summarizes the support for various type system features in ty.
...
+
+## Abstract base classes
+
+| Feature                                            | Status |
+| -------------------------------------------------- | ------ |
+| `@abstractmethod` decorator                        | ✅     |
+| Empty body allowed for abstract methods            | ✅     |
+| `@abstractmethod` with `@overload` validation      | ✅     |
+| Diagnostic: instantiating abstract class           | ❌     |
+| Diagnostic: missing abstract method implementation | ❌     |
```
AlexWaygood: How is this different to the previous one?

docs/reference/type-system-features.md
```diff
@@ -256,3 +392,34 @@ This page summarizes the support for various type system features in ty.
...
+| Non-method overridden by non-method          | ❌     |
+
+## Abstract base classes
+
+| Feature                                            | Status |
+| -------------------------------------------------- | ------ |
+| `@abstractmethod` decorator                        | ✅     |
+| Empty body allowed for abstract methods            | ✅     |
+| `@abstractmethod` with `@overload` validation      | ✅     |
+| Diagnostic: instantiating abstract class           | ❌     |
```
AlexWaygood: You can link https://github.com/astral-sh/ty/issues/1877

docs/reference/type-system-features.md
```diff
@@ -0,0 +1,425 @@
...
+| Diagnostic: missing abstract method implementation | ❌     |
+
+## `__slots__`
```
AlexWaygood, line 441: Could also mention that we currently only support slots if it's a single string or tuple of strings. We should also add support for list literals, dict literals and set literals

docs/reference/type-system-features.md
```diff
@@ -0,0 +1,425 @@
...
+| `hasattr()` narrowing                     | ✅                                                      |
+| `callable()` narrowing                    | ✅                                                      |
+| Assignment narrowing                      | ✅                                                      |
+| `TypeIs[…]` user-defined type guards      | ✅                                                      |
+| `TypeGuard[…]` user-defined type guards   | ❌ [#117](https://github.com/astral-sh/ty/issues/117)   |
+| `TypeIs`/`TypeGuard` as method            | ❌ [#1569](https://github.com/astral-sh/ty/issues/1569) |
+| Tuple length checks                       | ❌ [#560](https://github.com/astral-sh/ty/issues/560)   |
+| Tuple match case narrowing                | ❌ [#561](https://github.com/astral-sh/ty/issues/561)   |
+
+## Tuples
```
AlexWaygood, line 108: Could mention https://github.com/astral-sh/ty/issues/1741 as a limitation we have currently that other type checkers don't have

sharkdp:
Alex had a good idea: this is now a pinned issue (https://github.com/astral-sh/ty/issues/1889), which should hopefully reduce the maintenance overhead.

I will consider adding a link to this pinned issue from the documentation.

PR #1879, opened 2025-12-13
Add support for callout syntax (#1875)

### LINKED ISSUE #1875: Non-rendered callout in documentation
### Summary

GFM callout syntax is not rendered in mkdocs. it requires [markdown-callouts](https://github.com/oprypin/markdown-callouts#support-github-alerts-style-of-admonitions).

https://docs.astral.sh/ty/reference/configuration/#root_1

Image

### Version

ty 0.0.1-alpha.34

leandrobbraga:
This resolves the issue of Markdown callouts not rendering in MkDocs.

I used https://github.com/sondregronas/mkdocs-callouts, which is compatible with the syntax already being used in this specific part of the documentation.

## Before
image

## After
image

## Alternative Solution
An alternative solution would be to not add a new dependency and change the callout syntax in the original code. This was implemented in https://github.com/astral-sh/ruff/pull/21961.

```markdown
!!! warning "Deprecated"
    This option has been deprecated. Use `environment.root` instead.
```
image

Closes #1875

### DIFF
--- a/docs/requirements.in
+++ b/docs/requirements.in
@@ -8,3 +8,4 @@ mdformat-admon>=2.0.2
 mkdocs-redirects>=1.2.2
 mkdocs-git-revision-date-localized-plugin>=1.3.0
 mkdocs-llmstxt>=0.2.0
+mkdocs-callouts>=1.16.0
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -87,9 +87,12 @@ mergedeep==1.3.4
 mkdocs==1.6.0
     # via
     #   -r docs/requirements.in
+    #   mkdocs-callouts
     #   mkdocs-git-revision-date-localized-plugin
     #   mkdocs-material
     #   mkdocs-redirects
+mkdocs-callouts==1.16.0
+    # via -r docs/requirements.in
 mkdocs-get-deps==0.2.0
     # via mkdocs
 mkdocs-git-revision-date-localized-plugin==1.3.0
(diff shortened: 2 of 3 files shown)


MichaReiser:
Thank you. Do you know how this plugin is different from mkdocs-github-admonitions-plugin which we use in ruff

leandrobbraga:
My understanding is that they serve similar purposes: mkdocs-callout converts from Obsidian’s format, and mkdocs-github-admonitions-plugin converts from GitHub’s format.

The issue is that the syntax used is `> [!WARN] "Deprecated"`, which is compatible with Obsidian’s format but not GitHub’s. GitHub uses `WARNING`, not `WARN`, and it does not support titles such as "Deprecated."

I believe the better solution is to use the admonitions format directly: `!!! warning "Deprecated"`. This format is already used throughout the codebase and does not introduce a new dependency.

Some examples: 
- https://github.com/astral-sh/ruff/blob/82a7598aa8dc5cd2dde18e4da6e0e24b48dc7392/docs/editors/index.md?plain=1#L24
- https://github.com/astral-sh/ruff/blob/82a7598aa8dc5cd2dde18e4da6e0e24b48dc7392/docs/editors/settings.md?plain=1#L21
```markdown
!!! note "Added in Ruff `0.9.8`"

    The **Inline JSON configuration** option was introduced in Ruff `0.9.8`.
```

I’ve opened this alternative solution at https://github.com/astral-sh/ruff/pull/21961.

PR #1894, opened 2025-12-15
Restructure the content on the "Getting started" page

zanieb:
This restructures the existing "Getting started" content. I'll add more content in a separate pull request.

Note there will be a corresponding separate update to the README

PR #1899, opened 2025-12-15
Restructure and reduce the README content

zanieb:
Loosely corresponding to https://github.com/astral-sh/ty/pull/1894 but with more aggressive simplifications.

As with that pull request, this doesn't add a highlights section — I'll do that separately.

[Rendered](https://github.com/astral-sh/ty/tree/zb/ty-docs-readme?tab=readme-ov-file#ty)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Run ty with [uvx](https://docs.astral.sh/uv/guides/tools/#running-tools) to get 
 uvx ty check
 ```
 
-Or, check out the [online ty playground](https://play.ty.dev) to try it out on a snippet.
+Or, check out the [ty playground](https://play.ty.dev) to try it out in your browser.
 
 To learn more about using ty, see the [documentation](https://docs.astral.sh/ty/).
 


README.md
```diff
@@ -8,49 +8,31 @@ An extremely fast Python type checker and language server, written in Rust.
...
 ```shell
 uvx ty check
 ```
 
-ty will run on all Python files in the working directory and or subdirectories. If used from a
-project, ty will run on all Python files in the project (starting in the directory with the
-`pyproject.toml`)
-
-You can also provide specific paths to check:
+Or, check out the [online ty playground](https://play.ty.dev) to try it out on a snippet.
```
AlexWaygood: I sort-of prefer highlighting the playground first, actually (the way we currently do). You can have complex multi-file/multi-directory setups in the ty playground, not just snippets, and it lowers the burden to trying it out (no installation requried, you don't even need to switch from your browser to a terminal window).

AlexWaygood (approved):

README.md
zanieb: I think the playground is cool but realistically I think the vast majority of users are going to want to run ty locally first.

I mention snippets and small projects in the documentation section of the playground, but wanted to be succinct here.

AlexWaygood: Fair enough. Maybe just make this edit

```suggestion
Or, check out the [online ty playground](https://play.ty.dev).
```

then?

zanieb: hm. I sort of want to hint what kind of thing you can do in the playground. 

Maybe?

```suggestion
Or, check out the [ty playground](https://play.ty.dev) to try it out in your browser.
```

AlexWaygood: I like that!

PR #1907, opened 2025-12-15
Restore the ignore commentary for virtual environment diagnostics

zanieb:
I dropped this in https://github.com/astral-sh/ty/pull/1894/files#diff-b4d68dc855d0f9476d3f2ee343853bd21bf82ea9960d0cf06661baa244439dd6L52-L54

I think we should just fix this behavior in ty, but it seems appropriate to retain the hint until then (just rephrased on a different page)


zanieb:
cc @AlexWaygood

PR #1908, opened 2025-12-15
Add a diagnostic image to the README

zanieb:
[Rendered](https://github.com/astral-sh/ty/tree/zb/diagnostic-readme?tab=readme-ov-file#ty)


zanieb:
Idk about this one

sharkdp:
> Idk about this one

Me neither. This looks too much like an annoying error that I'd rather not like to see anywhere. And it has little to do with types.

zanieb:
I'll probably just use the benchmark but happy to try another diagnostic image if you are into that part.

PR #1909, opened 2025-12-15
Link to the typing specification in the "type checking" features page

zanieb:
Instead of to the top-level page which does not really seem like what I would want if I was trying to understand supported features?


sharkdp:
unsure. I'm okay either way.

zanieb:
I was caught off guard by the "tutorials" and "guides"

AlexWaygood:
The spec isn't really designed as end-user documentation — it's really targeted towards authors of type checkers — but it might well be the best thing to link to here anyway

AlexWaygood:
If we want better docs to link to I think we just have to write it ourselves, like mypy did

PR #1919, opened 2025-12-16
Minor updates to the docs

sharkdp:
## Summary

I did a full review of all docs using Claude, and these seemed worth fixing.

### DIFF
--- a/docs/suppression.md
+++ b/docs/suppression.md
@@ -19,7 +19,7 @@ Rule violations spanning multiple lines can be suppressed by adding the comment 
 violation's first or last line:
 
 ```py
-def sum_three_numbers(a: int, b: int, c: int): ...
+def sum_three_numbers(a: int, b: int, c: int) -> int: ...
 
 # on the first line
 


docs/suppression.md
```diff
@@ -91,12 +91,12 @@ to suppress all violations inside a function.
 ```python
 from typing import no_type_check
 
-def add_three(a: int, b: int, c: int):
-    a + b + c
+def add_three(a: int, b: int, c: int) -> int:
+    return a + b + c
```
sharkdp: I was confused when reading this example because I focused on the (apparently) wrong code in `add_three`, while I should be focusing on the wrong call-site in `main`. I guess this is just a Rust-vs-Python issue and this is what was actually meant.

docs/suppression.md
```diff
@@ -91,12 +91,12 @@ to suppress all violations inside a function.
...
 
 @no_type_check
 def main():
-    add_three(3, 4)
+    add_three(1, 2)  # no error for the missing argument
```
sharkdp: Changing arguments to `1, 2` to make it even more apparent that one is missing

docs/suppression.md
AlexWaygood: My intuition would be that an `add_three` function would take a single integer and add `3` to it. So we could also consider renaming it

```suggestion
def sum_three_ints(a: int, b: int, c: int) -> int:
    return a + b + c
```

(need to also update the use of the function a few lines below, if you accept this)

docs/type-checking.md
```diff
@@ -24,7 +24,7 @@ See the [module discovery](./modules.md) documentation for details.
 
 ## File selection
 
-ty will run on all Python files in the working directory and or subdirectories. If used from a
+ty will run on all Python files in the working directory and its subdirectories. If used from a
```
AlexWaygood: Maybe worth mentioning that it recurses into subdirectories of subdirectories?

```suggestion
ty will run on all Python files in the working directory and its (recursive) subdirectories. If used from a
```

PR #1924, opened 2025-12-16
pyproject.toml: Bump development status

sharkdp:
Looks like this is not something that rooster would do for us.


pyproject.toml
```diff
@@ -8,7 +8,7 @@ readme = "README.md"
 authors = [{ name = "Astral Software Inc.", email = "hey@astral.sh" }]
 keywords = ["ty", "typing", "analysis", "check"]
 classifiers = [
-    "Development Status :: 3 - Alpha",
+    "Development Status :: 4 - Beta",
```
sharkdp, line 11: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#classifiers

AlexWaygood (approved):

PR #1930, opened 2025-12-16
Add incrementality section to LSP docs

sharkdp:
I was halfway through with designing an example here to understand fine-grained incrementality, but ultimately, I'm not sure this is the right place to discuss this? It is interesting in a technical talk, but it's (ideally) really not relevant for users to understand this in detail(?).

### DIFF
--- a/docs/features/language-server.md
+++ b/docs/features/language-server.md
@@ -97,8 +97,8 @@ analyzed in context, with diagnostics, completions, and other features working a
 ty's architecture is designed for low-latency updates of diagnostics and other language server features.
 When you make a change in your editor, ty incrementally updates only the affected parts of the
 codebase, rather than re-analyzing everything from scratch. This happens at a fine-grained level,
-down to individual definitions. All of this is to ensure that you get instant feedback as you type (within
-a few milliseconds, even on large projects).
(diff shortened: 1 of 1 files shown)


docs/features/language-server.md
```diff
@@ -92,6 +92,19 @@ ty surfaces useful contextual information as you code:
 ty supports Jupyter notebooks (`.ipynb` files) with language server features. Each cell is
 analyzed in context, with diagnostics, completions, and other features working across cells.
 
+## Fine-grained incrementality
+
+ty's architecture is designed for low-latency updates of diagnostics and other language server features.
+When you make a change in your editor, ty incrementally updates only the affected parts of the
+codebase, rather than re-analyzing everything from scratch. This happens at a fine-grained level,
+down to individual definitions. All of this is to ensure that you get instant feedback as you type (within
+a few milliseconds, even on large projects).
```
zanieb:
```suggestion
down to individual definitions. This incrementality means that you get instant feedback as you type, i.e., within
a few milliseconds, even on large projects.
```

zanieb:
It might be nice to have a very simple example, but I agree it's not critical

PR #1944, opened 2025-12-16
Add documentation for `--watch` mode

zanieb:


docs/type-checking.md
```diff
@@ -40,6 +40,18 @@ ty check example.py
 
 You can also persistently configure [included and excluded files](./exclusions.md).
 
+## Watch mode
```
MichaReiser, line 52: Feels a bit weird to have it between file selection and rule selection. I would either move it directly after running the type checker or before type system

MichaReiser (approved):

Issue #2003, opened 2025-12-17
Add more documentation to special forms in `ty_extensions.pyi` stub


AlexWaygood:
It's easy for users to jump to these definitions in the stub file from inlay hints. It will be even easier to jump to these definitions if something like https://github.com/astral-sh/ty/issues/1575 is implemented.

We should add more docstrings/comments to the stub file so that users have a better experience when they land there. Ideally we would also show this documentation on hover, so that users can easily learn about what the `Unknown` type means and represents, for example.

Gankra:
(Also having those docs would let us show those docs whenever you hover a special form, which would be incredible)

sharkdp:
> (Also having those docs would let us show those docs whenever you hover a special form, which would be incredible)

Not directly, actually! I tried to implement this quickly before, but [since `Unknown` is not *declared*](https://github.com/astral-sh/ruff/blob/0bd7a94c2732c232cc1423756e2437d9bce558b1/crates/ty_vendored/ty_extensions/ty_extensions.pyi#L20), we show this:

<img width="785" height="265" alt="Image" src="https://github.com/user-attachments/assets/28c4ffd7-d2e0-4947-9076-558441f62fa5" />

AlexWaygood:
Yes I think three things are required for good on-hover with `Unknown`:
- Change the `Unknown` definition to `Unknown: _SpecialForm`. `Unknown = object()` is just really confusing because users get the docs for `object`
- Add an "attribute docstring" to `Unknown`, e.g.

  ```py
  Unknown: _SpecialForm
  """Some docs for `Unknown`"""
  ```

  (The PEP for these was [rejected](https://peps.python.org/pep-0224/), but they're nonetheless very popular in the community and recognised by the Sphinx documentation tool)
- Add support for showing attribute-docstrings on-hover to the LSP.

I think each of these stages on its own would be a big UX improvement, so we can do each of them in isolation

Gankra:
Item 3 is here, so y'all can go nuts with docs now:

* https://github.com/astral-sh/ruff/pull/22036

charliermarsh:
I _think_ this is sufficiently fixed by https://github.com/astral-sh/ruff/pull/26263.

Issue #585, opened 2025-06-05
Improve diagnostics (and documentation) for unresolved imports caused by editable installs that use setuptools as their build backend


SoftwareApe:
### Summary

Hi,

I hope this isn't a duplicate issue. I scanned through https://github.com/astral-sh/ty/issues/445 and the current open issues but didn't find it.

I have some development submodules added through `uv add --editable ...`. These are found in the `tool.uv.sources` section of the `pyproject.toml`.

They seem to work fine with uv and ruff. But `ty` complains that:

```
error[unresolved-import]: Cannot resolve imported module ...
```

PS: Amazing work from the astral team as always. This alpha version is already better than many mature python type-checkers!

### Version

`ty==0.0.1a8`

AlexWaygood:
This is probably a duplicate of https://github.com/astral-sh/ty/issues/475 if your development submodules are using `setuptools` for their build backend. But we've decided internally that we want to prioritise better diagnostics for this situation, so I'll keep this issue open to track that 👍

Thanks for the report!

SoftwareApe:
@AlexWaygood thank you so much! It indeed seems like #475 it's the same issue. I guess I didn't search for the right combination of strings. It's indeed a setuptools dependency.

carljm:
I've reopened #475 to see if there's any way we can support these -- but if we can't, we should still improve the diagnostics.

MichaReiser:
A first step (and very easy) could be to add some debug or info level logging when ty skips a `pth` file.

https://github.com/astral-sh/ruff/blob/64edfb6ef6f26bb3618c5c888f5abf3a866a172c/crates/ty_python_semantic/src/module_resolver/resolver.rs#L600-L615

MichaReiser:
Some notes from our internal discord on how we could detect packages that use setuptools

> Would you go and try to find a matching .pth file with wild python code in it? 
> and there are other things that setuptools inserts into site-packages that you can look for to know for sure that the package is installed, we just have no idea where the package actually is located
> and those things will also tell us that it was setuptools that installed it

carljm:
Some useful links for future reference: https://github.com/pypa/setuptools/issues/3518 and https://docs.basedpyright.com/v1.25.0/usage/import-resolution/#setuptools

mfish33:
[...]

Issue #487, opened 2025-05-22
cannot resolve lxml.etree


johnniemorrow:
### Summary

Hi ty-team,

I was trying to fix up some unresolved-attribute errors reported by ty related to lxml.etree in [lxml](https://pypi.org/project/lxml/).

Reducing it down to just an import statement, ty doesn't seem to be able to see lxml.etree:

```bash
$ cat use_lxml.py 
import lxml.etree

```

```bash
$ ty check use_lxml.py 
WARN ty is pre-release software and not ready for production use. Expect to encounter bugs, missing features, and fatal errors.
error[unresolved-import]: Cannot resolve imported module `lxml.etree`
 --> use_lxml.py:1:8
  |
1 | import lxml.etree
  |        ^^^^^^^^^^
  |
info: make sure your Python environment is properly configured: https://github.com/astral-sh/ty/blob/main/docs/README.md#python-environment
info: rule `unresolved-import` is enabled by default

Found 1 diagnostic

```

sharkdp:
Thank you for reporting this.

`lxml.etree` is a C extension module. I wonder if we could detect this and show a better error?

You probably want to fix this by installing `types-lxml`?

AlexWaygood:
> `lxml.etree` is a C extension module. I wonder if we could detect this and show a better error?

Pyright does a fancy thing in its module resolver to distinguish between modules that are definitely unresolvable (no python or C-extension modules are available) and modules which might be resolvable but which it cannot inspect, because it's a C extension or similar and there are no stubs installed.

Mypy does not have that feature, but it does have a hardcoded list of untyped packages for which it knows high-quality stubs are available, and issues a nice diagnostic suggesting to install the relevant stubs package if it issues an `unresolved-import` diagnostic on one of those packages.

We don't have either feature right now; we could consider adding one or both. (The mypy one is easier to implement!)

johnniemorrow:
> Thank you for reporting this.
> 
> `lxml.etree` is a C extension module. I wonder if we could detect this and show a better error?
> 
> You probably want to fix this by installing `types-lxml`?

`uv add types-lxml --dev` does the trick - thanks @sharkdp!

AlexWaygood:
#715 includes some discussion on exactly which extensions should be recognized when resolving modules to C extensions, and some other details of the implementation.

carljm:
#1613 raises the same scenario, and also suggests that we might be able to automatically introspect the `.so` and generate some implicit default stubs automatically.

[...]

PR #2006, opened 2025-12-17
Add typing FAQ

sharkdp:
Rendered: https://shark.fish/ty-faq/reference/typing-faq/

### DIFF
--- a/docs/reference/typing-faq.md
+++ b/docs/reference/typing-faq.md
@@ -116,7 +116,8 @@ ty reports *"Cannot resolve imported module …"*, check the following:
 
 1. **Compiled extensions**: ty requires `.py` or `.pyi` files for type information. If a package
     contains only compiled extensions (`.so` or `.pyd` files), you'll need stub files (`.pyi`) for ty
-    to understand the types.
+    to understand the types. See also [this issue](https://github.com/astral-sh/ty/issues/487) which
+    tracks improvements in this area.
 
 ## Does ty support monorepos?
 


docs/reference/typing-faq.md
```diff
@@ -0,0 +1,176 @@
...
+    [tool.ty.environment]
+    root = ["./app"]
+    ```
+
+1. **Third-party packages**: Ensure dependencies are installed in your virtual environment. Run ty
+    with `-v` to see the search paths being used.
+
+1. **Compiled extensions**: ty requires `.py` or `.pyi` files for type information. If a package
+    contains only compiled extensions (`.so` or `.pyd` files), you'll need stub files (`.pyi`) for ty
+    to understand the types.
```
sharkdp: This is not my area of expertise. Can someone proof-read and possibly extend this?

docs/reference/typing-faq.md
```diff
@@ -0,0 +1,176 @@
...
+1. **Configure multiple source roots**: Use [`environment.root`](./configuration.md#root) to specify
+    multiple source directories:
+
+    ```toml
+    [tool.ty.environment]
+    root = ["packages/package-a", "packages/package-b"]
+    ```
+
+    This has the disadvantage of treating all packages as a single project, which may lead to cases
+    in which ty thinks something is importable when it wouldn't be at runtime.
```
sharkdp, line 146: Similar here. Should we be advertising these workarounds?

docs/reference/typing-faq.md
AlexWaygood: This looks very solid to me. You could possibly link to https://github.com/astral-sh/ty/issues/487 here as the tracking issue for improving our diagnostics when C extensions are installed without stubs.

https://github.com/astral-sh/ty/issues/585 is also a very common issue (local package is installed editably, but using a dynamic setuptools editable install rather than a "traditional" editable install using a `.pth` file, so we can't see that the module is installed and can't resolve the import).

docs/reference/typing-faq.md
AlexWaygood: @zsol?

docs/reference/typing-faq.md
```diff
@@ -0,0 +1,176 @@
+# Typing FAQ
+
+This page answers some commonly asked questions about ty and Python's type system.
+
+## What is the `Unknown` type and when does it appear?
```
AlexWaygood, line 5: (https://github.com/astral-sh/ty/issues/2003 will also help a lot with confusion around this, I think)

docs/reference/typing-faq.md
sharkdp: > This looks very solid to me. You could possibly link to #487 here as the tracking issue for improving our diagnostics when C extensions are installed without stubs.

Done

> #585 is also a very common issue (local package is installed editably, but using a dynamic setuptools editable install rather than a "traditional" editable install using a `.pth` file, so we can't see that the module is installed and can't resolve the import).

I'm not sure I would do a good job trying to explain this. But that issue only has two upvotes, and one is from Carl, so I think it's fine to leave it out of the FAQs for now :smile:

docs/reference/typing-faq.md
sharkdp: Merging for now, we can always remove it later.

docs/reference/typing-faq.md
AlexWaygood: It's come up a bunch of times on discord too. But don't worry about it, I can add something as a followup!

docs/reference/typing-faq.md
zsol: It's not a big deal but I would probably not advertise the second option here - I can follow up with improvements once I've converted pyx to use `uv workspace` to iterate over members.

PR #2046, opened 2025-12-18
Bump version to 0.0.3

oconnor663:


docs/installation.md
```diff
@@ -71,7 +71,7 @@ ty includes a standalone installer.
     Request a specific version by including it in the URL:
 
     ```console
-    $ curl -LsSf https://astral.sh/ty/0.0.2/install.sh | sh
+    $ curl -LsSf https://astral.sh/ty/0.0.3/install.sh | sh
```
oconnor663, line 74: I'm not sure what to think about these changes (made by `release.sh`). Do we expect to change the install URL with every update going forward?

charliermarsh: Yeah, this looks right.

PR #2049, opened 2025-12-18
Create patch releases by default

charliermarsh:
## Summary

Removing `default-bump-type = "pre"` will tell Rooster to default to patch releases, now that the Beta is out.


carljm (approved):
So "patch" is the default bump type?

charliermarsh:
Yeah I believe so ([source](https://github.com/zanieb/rooster/blob/da1ac78d2fe5aa1926fe1a492eb657aa2d8143a7/src/rooster/_config.py#L56)). This matches what we use in `uv`.

zanieb:
Yep, thanks!

Issue #2038, opened 2025-12-17
error[unresolved-attribute]: Object of type `(...) -> Any` has no attribute `__name__`


kracekumar:
### Summary

The Python callable has attribute `__name__`but ty reports as unresolved-attribute.

Snippet

```python
def export(defn: Callable[..., Any]) -> Callable[..., Any]:
    """Decorator to explicitly mark functions that are exposed in a lib."""
    globals()[defn.__name__] = defn
    __all__.append(defn.__name__)
    return defn

```

Diagnostics

```
error[unresolved-attribute]: Object of type `(...) -> Any` has no attribute `__name__`
  --> litecli/packages/special/__init__.py:12:15
   |
10 | def export(defn: Callable[..., Any]) -> Callable[..., Any]:
11 |     """Decorator to explicitly mark functions that are exposed in a lib."""
12 |     globals()[defn.__name__] = defn
   |               ^^^^^^^^^^^^^
13 |     __all__.append(defn.__name__)
14 |     return defn
   |
info: rule `unresolved-attribute` is enabled by default
```

mypy: https://mypy-play.net/?mypy=latest&python=3.9&gist=8fca5c11fe239843ba39dc08aa0132ab
ty: https://play.ty.dev/01eb2208-8168-444c-a0b7-81bc20be124d

### Version

ty 0.0.2 (4283557 2025-12-16) Python: 3.9

carljm:
Thanks for the report! ty is alerting you to a real problem in your code here: not all callables have a `__name__` attribute. Functions and lambdas do, but there are other ways to create a callable in Python:

```pycon
>>> class C:
...     def __call__(self):
...         return None
...
>>> C().__name__
Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    C().__name__
AttributeError: 'C' object has no attribute '__name__'. Did you mean: '__ne__'?
```

You can make your code robust against this by checking `isinstance(defn, types.FunctionType)`, and ty will then be happy: https://play.ty.dev/725cbad9-14e7-4804-bf57-310cd118c217

There's further discussion in https://github.com/astral-sh/ty/issues/1495 of ways we could improve this experience in ty.

kracekumar:
Thank you for the link and suggestion. I think the `if` check wouldn't hurt.

kracekumar:
I ended up using `isinstance check of callabale and hasattr(defn, "__name__")`. 

```python
[...]

Issue #1858, opened 2025-12-11
support calls to intersection types


carljm:
Right now we infer a todo type, which suppresses false positives. But intersections arise reasonably frequently in narrowing, and often don't simplify (especially with negative or dynamic elements). So this can lead to a lot of false negatives.

The algorithm here isn't particularly hard. We should try to call each element of the intersection. We can probably short-circuit some cases (most negative intersection elements will simply fallback to `object`, and we know `object` is not callable). If calls to any element(s) succeed, the return type is the intersection of their returns, and we should ignore errors from those that failed. If all fail, we probably want to discard errors from some very general types (the same cases we considered short-circuiting above) and display the other errors. (Handling the diagnostics is probably the most complex piece of this issue.)

We could also do a very simple initial version here where we try to short-circuit negative elements and if that reduces the intersection to a single type, call it normally. That would already improve things a lot for narrowing cases.

AlexWaygood:
IIRC, according to https://jellezijlstra.github.io/negation-types.html we should just ignore negative elements altogether for most operations on intersection types?

carljm:
Yes, that's for the reason I mentioned: except for negation of some special very broad types like `AlwaysTruthy` and `AlwaysFalsy` in very particular scenarios (boolean checks), a negative intersection element generally can't be treated any differently from `object` -- and if it were `object` it would simplify out of the intersection anyway.

I think it would be a fairly easy and worthwhile experiment to just eagerly strip negative elements from intersection types altogether and see where that degrades our inference in practice.

AlexWaygood:
sorry, should have read more carefully :( blame it on the fever.

PR #2055, opened 2025-12-18
Typing FAQ: Attributes on Callable

sharkdp:
Add a new FAQ entry based on issues like https://github.com/astral-sh/ty/issues/2038

### DIFF
--- a/docs/reference/typing-faq.md
+++ b/docs/reference/typing-faq.md
@@ -91,9 +91,9 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
 
 When you access `__name__`, `__qualname__`, `__module__`, or `__doc__` on a value typed as `Callable`,
 ty reports an `unresolved-attribute` error. This is because not all callables have these attributes.
-Functions and lambdas do, but other callable objects do not. The `FileUpload` class below, for example,
-is callable, but does not have a `__name__` attribute. Passing an instance of `FileUpload` to `retry`
-would lead to an `AttributeError` at runtime.
+Functions do (including lambdas), but other callable objects do not. The `FileUpload` class below, for
+example, is callable, but instances of `FileUpload` do not have a `__name__` attribute. Passing a
+`FileUpload` instance to `retry` would lead to an `AttributeError` at runtime.
 
 ```py
 from typing import Callable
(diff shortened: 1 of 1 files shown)


docs/reference/typing-faq.md
```diff
@@ -87,6 +87,81 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
...
+    `types.FunctionType`:
+
+    ```py
+    from typing import Callable, TYPE_CHECKING
+    from types import FunctionType
+
+    if TYPE_CHECKING:
+        from ty_extensions import Intersection
+
+        type FunctionLikeCallable[**P, R] = Intersection[Callable[P, R], FunctionType]
```
sharkdp, line 154: So cool that this works now! I'd love to add this type alias to `ty_extensions`, but we should probably add support for calling intersection types first (#1858).

sharkdp: This is also a good example for advertising intersection types.

docs/reference/typing-faq.md
```diff
@@ -87,6 +87,81 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
...
+        raise NotImplementedError
+
+retry(3, FileUpload("image.png"))
+```
+
+To fix this, you could use a `hasattr(…, "__name__")` check to fall back to a default name when the
+attribute is not present:
+
+```py
+name = operation.__name__ if hasattr(operation, "__name__") else "operation"
```
spaceone:
```suggestion
To fix this, you could use a `getattr(…, "__name__", default)` check to fall back to a default name when the
attribute is not present:

```py
name = getattr(operation, "__name__", "operation")
```

docs/reference/typing-faq.md
```diff
@@ -87,6 +87,81 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
 
     If you need this for `complex`, you can use `ty_extensions.JustComplex` in a similar way.
 
+## Why does ty say `Callable` has no attribute `__name__`?
+
+When you access `__name__`, `__qualname__`, `__module__`, or `__doc__` on a value typed as `Callable`,
+ty reports an `unresolved-attribute` error. This is because not all callables have these attributes.
+Functions and lambdas do, but other callable objects do not. The `FileUpload` class below, for example,
```
AlexWaygood:
```suggestion
Functions do (including lambdas), but other callable objects do not. The `FileUpload` class below, for example,
```

(`lambda`s are instances of `FunctionType` at runtime, just like functions created using `def` statements)

docs/reference/typing-faq.md
```diff
@@ -87,6 +87,81 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
...
+is callable, but does not have a `__name__` attribute. Passing an instance of `FileUpload` to `retry`
```
AlexWaygood:
```suggestion
is callable, but instances of `FileUpload` do not have a `__name__` attribute. Passing a `FileUpload` instance to `retry`
```

docs/reference/typing-faq.md
```diff
@@ -87,6 +87,81 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
...
+        if operation():
+            return True
+    return False
+
+class FileUpload:
+    def __init__(self, name: str) -> None:
+        raise NotImplementedError
+
+    def __call__(self) -> bool:
+        raise NotImplementedError
```
AlexWaygood: I think it's easier for users to understand if we fill in the bodies here, or use `...` to visually indicate that "something's being left out"

```suggestion
        ...

    def __call__(self) -> bool:
        return True
```

docs/reference/typing-faq.md
AlexWaygood: I agree that using `getattr()` with three arguments is probably more idiomatic here

Issue #2001, opened 2025-12-17
Rename `invalid-method-override` rule


AlexWaygood:
We have lots of rules that complain about invalid method overrides of one sort or another, and we'll be adding more in the future. The `invalid-method-override` rule is specifically about Liskov violations, but its name doesn't reflect that currently. `unsound-method-override` might be better, though it's still not great. Maybe `incompatible-method-override`?

Issue #1884, opened 2025-12-14
Subheadings not rendered in `invalid-method-override` documentation


AlexWaygood:
Our [docs for `invalid-method-override`](https://docs.astral.sh/ty/reference/rules/#invalid-method-override) are longer than the docs for most of our rules. I wanted to make them high-quality, because I expect that many users won't have heard of the Liskov Substitution Principle, and it's useful to have a good explanation of why it's important for us to link to.

As part of the documentation, I added a ["Common issues" section](https://github.com/astral-sh/ruff/blob/8bc753b842967b53dc2698fdff8e1a410ffafa46/crates/ty_python_semantic/src/types/diagnostic.rs#L2170-L2188), with the idea that I'd be able to link directly to it or subsections in it if users brought up a question that we'd already written up an answer to. Currently, however, the `## Common issues` heading is the same font size as the `### Why does ty complain about my `__eq__` method?` heading immediately below it. I also can't link directly to any subsections in the docs for this rule currently.

We should either find a way to fix this, or I should rewrite the docs for this rule to not use subheadings 🙃

dhruvmanila:
I think it might be useful to have separate pages for each rule (similar to Ruff) which would allow us to provide links to individual sections in a rule documentation. We'd need to have a design on how to present a list of rules similar to https://rust-lang.github.io/rust-clippy/master/index.html and add some filters / sorting ability based on attributes like severity, version added, etc.

PR #2126, opened 2025-12-20
two new typing FAQ entries

carljm:
Add new FAQ entries about checking rule code docs, and `Top` types.

### DIFF
--- a/docs/reference/typing-faq.md
+++ b/docs/reference/typing-faq.md
@@ -2,6 +2,11 @@
 
 This page answers some commonly asked questions about ty and Python's type system.
 
+## Why does ty report an error on my code?
+
+Check the [documentation](https://docs.astral.sh/ty/reference/rules/) for the specific error code
+you are seeing; it may explain the problem.
+
 ## What is the `Unknown` type and when does it appear?
 
 `Unknown` is ty's way of representing a type that could not be fully inferred. It behaves the same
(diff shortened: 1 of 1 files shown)


docs/reference/typing-faq.md
```diff
@@ -251,3 +251,53 @@ No. ty does not have a plugin system and there is currently no plan to add one.
...
+
+This type represents "all possible lists of any element type" (as opposed to `list[Unknown]`, which
+represents "a list of some unknown element type"). It usually arises from a check such as
+`if isinstance(x, list):`. If `x` was previously of type `Item | list[Item]`, you may expect this check
+to narrow the type to `list[Item]`, but ty respects the possibility that there could be a common
+subclass of both `Item` and `list` (which may not be a list of `Item`!), and so the narrowed type is
+instead `(Item & Top[list[Unknown]]) | list[Item]`. This code can be made more robust by instead
+checking `if instance(x, Item)`, or by declaring the `Item` type as `@typing.final`.
+
+See also [discussion
```
sharkdp: Maybe
```suggestion
See also the [discussion
```

docs/reference/typing-faq.md
```diff
@@ -251,3 +251,53 @@ No. ty does not have a plugin system and there is currently no plan to add one.
 We prefer extending the type system with well-specified features rather than relying on
 type-checker-specific plugins. That said, we are considering adding support for popular third-party
 libraries like pydantic, SQLAlchemy, attrs, or django directly into ty.
+
+## Why does ty say I have an invalid method override?
```
AlexWaygood: I already have a mini-FAQ for this rule in the docs for this rule at https://docs.astral.sh/ty/reference/rules/#invalid-method-override, under the "Common issues" subheading. That isn't rendered very well right now (https://github.com/astral-sh/ty/issues/1884), but I think it's useful for all the docs relating to this rule to be in one place. Another advantage of putting FAQs there is that we link to that page straight from diagnostics printed to the terminal (following https://github.com/astral-sh/ruff/pull/21502).

So I would be inclined to improve the FAQ section at https://docs.astral.sh/ty/reference/rules/#invalid-method-override, and link to that page from this document, rather than having Liskov-related FAQs in both documents.

carljm: How do we feel about linking to open GH issues from rules docs?

carljm: > improve the FAQ section

This will be a separate PR, since it's a separate repo.

> and link to that page from this document

I added an FAQ here that does this in a very generic way, for all rule codes -- lmk what you think.

AlexWaygood: > How do we feel about linking to open GH issues from rules docs?

I think that's fine? Doesn't feel materially different to me to linking to GH issues from the FAQs doc here

zanieb: We don't do it a lot, but I do think it's helpful in cases. I usually put it in an admonition.

docs/reference/typing-faq.md
```diff
@@ -2,6 +2,11 @@
 
 This page answers some commonly asked questions about ty and Python's type system.
 
+## Why does ty report an error on my code?
+
+Check the [documentation](https://docs.astral.sh/ty/reference/rules/) for the specific error code
+you are seeing; it may explain the problem.
```
AlexWaygood, line 8: From a user perspective, it would probably be nicer to also have direct links to the error codes where we know we have different behaviour to mypy/pyright -- something like:

> Rules where we have slightly different behaviour to mypy and pyright include:
> - `invalid-method-override`
> - `foo-bar-baz`
> - `spam-ham-eggs`

But the advantage of the way you've currently got it is that we won't have to remember to change the link here if we rename a rule in the future (https://github.com/astral-sh/ty/issues/2001). So: no strong opinion.

PR #2145, opened 2025-12-21
[docs] Switch override and src order in configurations

Cjkjvfnby:
## Summary

It's confusing that 'override' is above the 'src' section.  Typically, you begin by configuring the global settings and then fine-tune them.

## Test Plan

No test is required for docs.  I just switched the blocks of the document.

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -194,114 +194,6 @@ typeshed = "/path/to/custom/typeshed"
 
 ---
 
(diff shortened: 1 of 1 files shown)


zanieb:
This file is generated from the sources in the ruff repository.

Cjkjvfnby:
> This file is generated from the sources in the ruff repository.

Missed the header in the file.  Will create a review there.

PR #2167, opened 2025-12-22
Document `configuration` and `configurationFile` editor settings

MichaReiser:
Screenshot 2025-12-23 at 10 06 58

### DIFF
--- a/docs/reference/editor-settings.md
+++ b/docs/reference/editor-settings.md
@@ -8,7 +8,7 @@ Code extension][ty-vscode].
 In-editor configuration of ty's settings. The inline settings always take precedence over the settings from configuration files,
 including the configuration specified with [`configurationFile`](#configurationfile).
 
-See [configuration](../configuration.md) for the supported configuration options.
+Consult [the configuration reference](../configuration.md) for a list of all supported configuration options.
 
 **Default value**: `null`
 


docs/reference/editor-settings.md
```diff
@@ -3,6 +3,136 @@
...
+
+    ```lua
+    require('lspconfig').ty.setup({
+      settings = {
+        ty = {
+          configuration = {
+            rules = {
+              ["unresolved-reference"] = "warn"
+            }
+          }
```
MichaReiser, line 41: I hope I go the syntax correct

dhruvmanila: Yes, this is correct

docs/reference/editor-settings.md
```diff
@@ -3,6 +3,136 @@
 The editor settings supported by ty's language server, as well as the settings specific to [ty's VS
 Code extension][ty-vscode].
 
+## `configuration`
+
+In-editor configuration of ty's settings. The inline settings always take precedence over the settings from configuration files.
+See [configuration](../configuration.md) for the supported configuration options.
```
dhruvmanila: We should expand this to make it clear that the "configuration files" also include the ones provided in `configurationFile`

docs/reference/editor-settings.md
```diff
@@ -3,6 +3,136 @@
...
+
+**Default value**: `null`
+
+**Type**: `object`
+
+**Example usage**:
```
dhruvmanila, line 17: It would be cool to include these example usages in https://docs.astral.sh/ty/reference/configuration as well to answer "how to specify these settings directly in the editor?". We can create a follow-up issue for this

docs/reference/editor-settings.md
```diff
@@ -3,6 +3,136 @@
...
+    ```
+
+## `configurationFile`
+
+The path to a `ty.toml` configuration file. ty will use the specified configuration over any automatically discovered configuration.
+ty will expand a tilde `~` at the start of a string to the user's home directory, as well as variables like `$A` or `${A}`.
+
+!!! info
+
+While ty configuration can be included in a `pyproject.toml` file, it is not allowed in this context.
```
dhruvmanila: I believe this needs to be indented?


```suggestion
!!! info

    While ty configuration can be included in a `pyproject.toml` file, it is not allowed in this context.
```

docs/reference/editor-settings.md
dhruvmanila, line 1: It might be useful to provide a "Resolution order" between `configuration`, `configurationFile` and a local configuration file similar to https://docs.astral.sh/ruff/editors/settings/#configuration_resolution_order. And, I think in the future it would also include settings that can be directly specified.

MichaReiser: Agree, I intentionally skipped this for now to have something working sooner and get some user feedback too

docs/reference/editor-settings.md
MichaReiser: Not sure, it feels a bit annoying if we have to mention this for every setting.

docs/reference/editor-settings.md
MichaReiser: Hmm yes. Pre-commit reverts a single tab indent, but using two seems to work fine

PR #2231, opened 2025-12-26
update_schemastore.py: Fix actual vs expected revision in log

paarth-a:
## Summary
 Fix the expected vs actual wording in scripts/update_schemastore.py so the log now reports the commit recorded on main as the expected revision and the checked-out submodule commit as the actual revision.


MichaReiser:
thank you

PR #2252, opened 2025-12-29
Add mise usage instructions to installation guide

kvokka:
The tool was added to [mise](https://github.com/jdx/mise) registry with https://github.com/jdx/mise/pull/7495 , and this commit updates the docs

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -71,7 +71,7 @@ ty includes a standalone installer.
     Request a specific version by including it in the URL:
 
     ```console
-    $ curl -LsSf https://astral.sh/ty/0.0.7/install.sh | sh
+    $ curl -LsSf https://astral.sh/ty/0.0.10/install.sh | sh
     ```
 
 === "Windows"
@@ -87,7 +87,7 @@ ty includes a standalone installer.
     Request a specific version by including it in the URL:
 
     ```pwsh-session
-    PS> powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/ty/0.0.7/install.ps1 | iex"
+    PS> powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/ty/0.0.10/install.ps1 | iex"
     ```
 
 !!! tip
@@ -163,7 +163,7 @@ COPY --from=ghcr.io/astral-sh/ty:latest /ty /bin/
 The following tags are available:
 
 - `ghcr.io/astral-sh/ty:latest`
-- `ghcr.io/astral-sh/ty:{major}.{minor}.{patch}`, e.g., `ghcr.io/astral-sh/ty:0.0.7`
+- `ghcr.io/astral-sh/ty:{major}.{minor}.{patch}`, e.g., `ghcr.io/astral-sh/ty:0.0.10`
 - `ghcr.io/astral-sh/ty:{major}.{minor}`, e.g., `ghcr.io/astral-sh/ty:0.0` (the latest patch
     version)
 
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -40,7 +40,7 @@ jobs:
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-build') }}
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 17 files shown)


docs/installation.md
```diff
@@ -50,6 +50,21 @@ To update ty, use `uv tool upgrade`:
 uv tool upgrade ty
 ```
 
+### Using mise
```
zanieb: If we include this (which I think will require consensus from other maintainers too), I think it should go below all the official distribution methods instead of here. The title should also probably be consistent? e.g., "Installing with mise"?

docs/installation.md
```diff
@@ -50,6 +50,21 @@ To update ty, use `uv tool upgrade`:
...
+```shell
+# List available versions
+mise ls-remote ty
+
+# Install the latest version
+mise install ty
+
+# Use a specific version for the project
+mise use ty@0.0.7
+```
```
zanieb: Having multiple instructions in a single block doesn't match the other examples on this page.

docs/installation.md
```diff
@@ -50,6 +50,21 @@ To update ty, use `uv tool upgrade`:
 uv tool upgrade ty
 ```
 
+### Using mise
+
+For dependency management with [mise](https://github.com/jdx/mise):
+
+```shell
+# List available versions
+mise ls-remote ty
```
zanieb: I don't think we need to include this, we don't explain how to list versions for any of the other installers.

docs/installation.md
MichaReiser: I like how it's done in uv https://docs.astral.sh/uv/getting-started/installation/#homebrew or we could even have a single headline, `Or your package manager of choice` (or similar) and have all of them listed ina single section.

kvokka:
Made this one consistent with other installation methods.
But it feels like the question of how to organise the docs in terms of the different installation methods is a matter for another PR

docs/installation.md
```diff
@@ -50,6 +50,18 @@ To update ty, use `uv tool upgrade`:
 uv tool upgrade ty
 ```
 
+### Installing globally with mise
+
```
MichaReiser: I think this should come after `pipx`, but definitely after the standalone installer.

kvokka: Moved

docs/installation.md
```diff
@@ -130,6 +130,18 @@ To update ty, use `pipx upgrade`:
...
+Install ty globally with with [mise](https://github.com/jdx/mise):
+
+```shell
+# Install the latest version
+mise install ty
+
+# Set it globally
+mise use --global ty
```
MichaReiser, line 152: Should this use `mise use --global ty` instead (I don't know mise but that's what i found in the documentation)

docs/installation.md
```diff
@@ -138,6 +138,18 @@ Install ty into your current Python environment with pip:
...
+```
```
dhruvmanila, line 153: I think it would be useful to have these two commands in separate code blocks so that it's easier for the user to directly copy-paste it using the square icon on the top-right corner in the rendered documentation. And, the text can be moved outside:


````suggestion
```shell
mise install ty
```

To set it globally:

```shell
mise use --global ty
```
````

File: docs/installation.md [hunk 7]
```diff
@@ -138,6 +138,20 @@ Install ty into your current Python environment with pip:
...
 
+### Installing globally with mise
+
+Install ty globally with with [mise](https://github.com/jdx/mise):
+
+```shell
+mise install ty
+```
+
+To set it globally:
```
carljm, line 149: Sorry for yet another reviewer coming in late -- but I don't understand what "To set it globally" means, or how it differs from "Install ty globally" above?

kvokka: With `mise`, you install something once, and then it might be linked to some project or be the default choice. So the user set the installed tool for the environment.

In contrast, with tools like npm, the package will be installed for the specific project (or globally), keeping multiple copies on the disk. 

Hope that this explains the chosen wording

carljm: Works for me -- I assume it will be clear enough to a mise user.

carljm:
This already has an approving review, and feedback has been addressed; going ahead and merging it.

PR #2302, opened 2026-01-02
Fix `--no-respect-ignore-files` typo in CLI documentation

evroon:
## Summary
`ty check --help` returns this:

```
File selection:
      --respect-ignore-files
          Respect file exclusions via `.gitignore` and other standard ignore files. Use `--no-respect-gitignore` to disable
```

However, the option `--no-respect-gitignore` does not exist, it should be `--no-respect-ignore-files` instead.

## Test Plan

I am not sure testing is needed? I only fixed a typo


AlexWaygood:
Thanks! Unfortunately this file is generated from code in the ruff repo.

And it looks like this was fixed a couple of days ago, in https://github.com/astral-sh/ruff/commit/4f2529f353e3ec2f9382a0258952c9c58b08ce89! So this should be fixed automatically when we next run the release script to regenerate this file and cut a new release.

Thank you!

PR #2385, opened 2026-01-07
add emergency manual release script

Gankra:
Same one in uv but with the prerelease line delete because @zanieb noted it was broken

### DIFF
--- a/scripts/manual-github-release.sh
+++ b/scripts/manual-github-release.sh
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+#
+# Manually perform a GitHub release for a broken automated release that otherwise went out.
+#
+# PLEASE USE WITH CAUTION.
+#
+# It can take a while to download all the artifacts.
+#
+# Requires the `gh` CLI.
+
+set -euo pipefail
+
+if [ ! -n "$COMMIT" ]; then
+    echo "COMMIT is required."
+    exit 1
+fi
+
+if [ ! -n "$RUN_ID" ]; then
+    echo "RUN_ID is required."
+    exit 1
+fi
+
+# Create directory for artifacts
+mkdir -p "release_$RUN_ID"
+cd "release_$RUN_ID"
+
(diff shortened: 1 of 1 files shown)


scripts/manual-github-release.sh
```diff
@@ -0,0 +1,55 @@
...
+# Download all artifacts for the workflow run
+gh run download "$RUN_ID" --repo "$REPO" --pattern 'artifacts-*'
+
+MANIFEST="artifacts-dist-manifest/dist-manifest.json"
+
+# Extract values from manifest
+TAG=$(jq -r '.announcement_tag // .tag' "$MANIFEST")
+TITLE=$(jq -r '.announcement_title' "$MANIFEST")
+BODY=$(jq -r '.announcement_github_body' "$MANIFEST")
+PRERELEASE=$(jq -r '.announcement_is_prerelease' "$MANIFEST")
```
zanieb, line 38: This might be a problematic line too? I'm not sure we should omit it unless we reproduce the problem though. Also, I think no part of this script is repo specific, so we should probably just put it in it's own repo or a gist and link to it in CONTRIBUTING?

Gankra: I *think* it was just the later line with the echo that was messed up

MichaReiser: Regardless of where we add it. Can we add a mention to `CONTRIBUTING.md` that this script exist and when/how to use it

AlexWaygood:
We have the shellcheck hook enabled in Ruff's pre-commit config but not in this repo — it might be an idea to add it here too if we're adding more shell scripts. It's really good at checking for syntax issues and common mistakes in shell scripts. https://github.com/astral-sh/ruff/blob/7319c37f4eb063e9590e1f09c8e92d7dabc63403/.pre-commit-config.yaml#L134

charliermarsh:
I think we should just close this for now; we can revive it in the event that we need to proceed with a manual release in the future.

PR #2381, opened 2026-01-07
Bump version to 0.0.10

Gankra:

PR #2386, opened 2026-01-07
Revert "Bump version to 0.0.10 (#2381)"

Gankra:
This reverts commit d18902cdcc4d39badfad86c88e89b866a809c881.


carljm:
Not sure I followed all the discussion on this, but is this what we actually want to do? Seems like we got the wheels all uploaded and 0.0.10 is live on PyPI now.

charliermarsh:
I think the purpose here was to assess whether the PPC64 wheel was built with or without a manylinux tag after rolling back to v0.0.9 (to determine whether the issue was in our code or an unpinned dependency).

Gankra:
And indeed it was in an unpinned dependency

PR #2387, opened 2026-01-07
pin maturin version to the same one uv uses

Gankra:


zanieb:
We should look into automating bump of these. I can take that.

PR #2388, opened 2026-01-07
Fail if maturin produces a PyPI incompatible wheel

zanieb:
When we use automatic manylinux tag determination, we need to ensure that the wheel is uploadable to PyPI or the job will fail downstream and cause a partial release.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -297,7 +297,7 @@ jobs:
           - target: arm-unknown-linux-musleabihf
             arch: arm
             # This target produces a `linux_armv6l` wheel which PyPI accepts,
-            # but maturin's `--compatibility pypi` flag doesn't recognize it.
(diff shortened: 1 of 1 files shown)


.github/workflows/build-binaries.yml
```diff
@@ -296,6 +296,9 @@ jobs:
             maturin_docker_options: -e JEMALLOC_SYS_WITH_LG_PAGE=16
           - target: arm-unknown-linux-musleabihf
             arch: arm
+            # This target produces a `linux_armv6l` wheel which PyPI accepts,
+            # but maturin's `--compatibility pypi` flag doesn't recognize it.
```
zanieb:
```suggestion
            # This target produces a `linux_armv6l` wheel which PyPI accepts,
            # but maturin's `--compatibility pypi` flag rejects.
```

zanieb:
I'll let @konstin follow-up here before merging and porting elsewhere

PR #2389, opened 2026-01-07
Set the manylinux tag explicitly for ppc64 builds

zanieb:
As of the latest version of maturin, this actually may not be compatible with the manylinux tag, but I think we should continue publishing it until we've made the actual choice to drop support. This is redundant with #2387 which downgrades maturin, but relevant for future upgrades.


charliermarsh:
Does this _fail_ on the latest maturin version? (Like, requesting a tag that maturin views as incompatible?)

zanieb:
I don't think so, but I could put up a branch on top to check?

zanieb:
Superseded by https://github.com/astral-sh/ty/pull/2393

PR #2407, opened 2026-01-09
Declare support for Python 3.14

dhruvmanila:
## Summary

This PR officially declare support for Python 3.14 which I think we do looking at the changelog of 3.14.

Reference: https://docs.python.org/3/whatsnew/3.14.html

- ty always defers annotation on 3.14
- t-strings are supported
- `except` and `except *` are supported without parentheses
- What about https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-finally-syntaxwarning ?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,7 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
     "Programming Language :: Python :: 3 :: Only",
     "Programming Language :: Rust",
     "Topic :: Software Development :: Quality Assurance",


MichaReiser:
> What about [docs.python.org/3/whatsnew/3.14.html#whatsnew314-finally-syntaxwarning](https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-finally-syntaxwarning) ?

I don't think this is implemented, see https://play.ruff.rs/fb66ac1a-155e-49f8-9e1b-5dfae415d4da. But I also don't consider syntax warnings as blocking to claim support for a newer Python version.

MichaReiser (approved):

carljm:
Also I think adding that syntax warning was a mistake, and I would be fine if ty never emulated it 😆 (EDIT: more seriously, I think it would be a great lint rule for ty in future -- but it shouldn't be a runtime warning.)

Issue #2322, opened 2026-01-04
Make some background for the benchmark image


chirizxc:
https://pypi.org/project/ty/

If we go to PyPi, we can't see any text in the benchmark at all

<img width="905" height="557" alt="Image" src="https://github.com/user-attachments/assets/10de895c-d25b-438a-8214-e59313bcf6be" />

with [Dark Reader](https://darkreader.org/):

<img width="880" height="710" alt="Image" src="https://github.com/user-attachments/assets/49716006-9f2c-4cc8-a873-58df6dface98" />

I think if the image is generated with a white background, it will look better.

chirizxc:
It seems that this photo is also used in the documentation, in which case two photos will need to be generated🤔

MichaReiser:
We solved this in ruff (and I think also uv) by rewriting the README before publishing, see 
https://github.com/astral-sh/ruff/blob/388658efdb088e517bb127dce3f016c976cdd273/scripts/transform_readme.py#L7-L44

We should update ty's `transform_readme.py` and add a similar transformation step.

zanieb:
I can't reproduce this in Firefox or Vivaldi. The image has black text, not white? 

edit: Clarified in https://github.com/astral-sh/ty/pull/2422#issuecomment-3730282665

PR #2422, opened 2026-01-09
Remove dark-mode handling from PyPI-uploaded README

charliermarsh:
## Summary

This matches the approach we use in Ruff.

Closes https://github.com/astral-sh/ty/issues/2322.


zanieb:
Wait what? I thought I tested this by just using a light-mode image and GitHub automatically switches the text rendering in dark mode.

If I go to PyPI, I can see the text fine?

Screenshot 2026-01-09 at 1 22 55 PM

charliermarsh:
You can't see it if you enable dark mode:

Screenshot 2026-01-09 at 2 24 57 PM

charliermarsh:
Because our media query will switch the text to white, but the rest of the PyPI page doesn't adapt.

zanieb:
Note I intentionally removed the complicated handling from Ruff because it wasn't necessary, so please wait to merge this.

zanieb:
There is no media query though? It's just `Shows a bar chart with benchmark results.`?

You're referring to the embedded query?

zanieb:
You're saying this occurs if you visit PyPI with dark mode enabled?

charliermarsh:
> You're saying this occurs if you visit PyPI with dark mode enabled?

Yes.

> You're referring to the embedded query?

Yeah, there's CSS _within_ `ty-benchmark-cli.svg`. I removed it in this PR.

zanieb:
Thanks for clarifying. I missed https://github.com/astral-sh/ty/pull/2422#issuecomment-3730282665

zanieb:
I think I'd prefer to retain the adaptive image and add a static light image and just replace the single URL on publish? Then we don't need a srcset and don't need to replicate that in the replacement code?

charliermarsh:
That's fine with me, I'll do a pass on it.

zanieb:
Thank you!

PR #2423, opened 2026-01-09
Add `language: golang` to actionlint pre-commit hook

mschoettle:
## Summary

Add `language: golang` to the pre-commit hook of actionlint. That way, Renovate will update the shellcheck version in `additional_dependencies` in the future (x-ref https://github.com/astral-sh/ty/pull/2262).

See: https://docs.renovatebot.com/modules/manager/pre-commit/#go

## Test Plan

`uvx prek run --all-files --hook-stage=manual`


AlexWaygood:
Thank you very much!

PR #2459, opened 2026-01-12
Fix the docs links to the benchmark results

Iain-S:
## Summary

This PR corrects the HTML `<src=...>` attributes in `index.md` such that they point to the right SVG files.

Note:

1. There is a bar chart on the docs home page <https://docs.astral.sh/ty/>.
2. There are two SVG files and we choose which one to display depending on the theme (light or dark).
3. We currently get:
    1. `ty-benchmark-cli.svg`, if we use the light theme. This has light writing, which cannot be seen against the light theme background.
    2. `ty-benchmark-cli-dark.svg`, if we use the dark theme. This does not exist.
4. This PR changes these links so that we use `ty-benchmark-cli-light.svg` for the light theme and `ty-benchmark-cli.svg` for the dark theme. Both of these images already existed.

## Test Plan

We manually compare the before and after appearance of the docs site.

### Light

| before | after |
| - | - | 
| Screenshot 2026-01-12 at 10 17 27 | Screenshot 2026-01-12 at 10 21 47 |

### Dark

| before | after |
| - | - | 
| Screenshot 2026-01-12 at 10 17 19 | Screenshot 2026-01-12 at 10 21 43 |

## To the reviewer

The `CONTRIBUTING.md` guidelines recommend running

```shell
npx prettier --prose-wrap always --write "**/*.md"
```

after editing any markdown files. However, this changes 24 files, not only `index.md`. Should I run this as a separate commit, a separate PR or neither?

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,4 +1,4 @@
-# Configuration for the actionlint tool, which we run via pre-commit
+# Configuration for the actionlint tool, which we run via prek
 # to verify the correctness of the syntax in our GitHub Actions workflows.
 
 self-hosted-runner:
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -43,9 +43,9 @@
       enabled: false,
     },
     {
-      groupName: "pre-commit dependencies",
+      groupName: "prek dependencies",
       matchManagers: ["pre-commit"],
-      description: "Weekly update of pre-commit dependencies",
+      description: "Weekly update of prek dependencies",
     }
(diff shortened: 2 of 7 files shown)


dhruvmanila (approved):
Thank you!

dhruvmanila:
> However, this changes 24 files, not only `index.md`. Should I run this as a separate commit, a separate PR or neither?

Let's leave that for now, I'm not sure about that as I don't think we enforce it in the CI (cc @zanieb)

PR #2556, opened 2026-01-18
docs: Add beta status notice to README

### LINKED ISSUE #2462: Clarify beta status of ty
Hi,

It’s currently a bit unclear whether ty should be considered a beta release.

The [latest blog post](https://astral.sh/blog/ty) announcing ty explicitly refers to it as a beta:

> Today, we're announcing the Beta release of [ty](https://github.com/astral-sh/ty).

Similarly, the `pyproject.toml` includes a [beta classifier](https://github.com/astral-sh/ty/blob/6b5938a5cbff201f1a6bebad991df9095ee76cbf/pyproject.toml#L11).

However, the published version number (currently 0.0.11) does not include a beta pre-release segment as defined by [PEP 440](https://peps.python.org/pep-0440/#pre-releases), and the `README.md` does not mention a beta status.

It would be helpful to clarify the intended release status of ty and ensure it is communicated consistently—either by clearly marking it as beta everywhere, or by removing the beta designation entirely.

Thanks!

kirankotari:
Clarifies that ty is currently in beta and explains the 0.0.x version numbering.
This addresses issue #2462 by providing clear communication about the project's
beta status directly in the README.

Closes #2462

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,11 +21,7 @@ An extremely fast Python type checker and language server, written in Rust.
 ty is backed by [Astral](https://astral.sh), the creators of
 [uv](https://github.com/astral-sh/uv) and [Ruff](https://github.com/astral-sh/ruff).
 
-> [!NOTE]
-> ty is currently in **beta**. The `0.0.x` version numbering indicates that the
-> project has not yet reached a stable release. While ty is production-ready and
-> actively maintained, users should expect some changes as the project evolves
-> toward its first stable release.
+ty is currently in [beta](#version-policy).
 
 ## Highlights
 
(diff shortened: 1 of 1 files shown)


MichaReiser:
Thank you. This feels to prominent to me. 

I think what I would do here is to add a `Version Policy` section and explain there that ty is still under active development and, while we're mindful about breaking changes, these can occur between any two patch versions.

eitan-protego:
It would be very helpful to have a prominent notice with a link to an implementation status doc - it's very hard to make decisions about whether to adopt ty if it's not obvious what type system features are and aren't implemented.

PR #2571, opened 2026-01-20
Drop PPC64 (big endian) builds

konstin:
PPC64 seems dead, and it's only supported on one exact manylinux version (https://github.com/pypa/auditwheel/issues/669), so we should drop it.

It's arguable whether this is a breaking change, since it doesn't remove support for the platform (which is already barely support by manylinux), but it does make installation more complex on PPC64.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -299,11 +299,6 @@ jobs:
             manylinux: 2_17
             # see https://github.com/astral-sh/ruff/issues/10073
             maturin_docker_options: -e JEMALLOC_SYS_WITH_LG_PAGE=16
-          - target: powerpc64-unknown-linux-gnu
-            arch: ppc64
-            manylinux: 2_17
-            # see https://github.com/astral-sh/ruff/issues/10073
-            maturin_docker_options: -e JEMALLOC_SYS_WITH_LG_PAGE=16
           - target: arm-unknown-linux-musleabihf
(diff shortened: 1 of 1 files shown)


carljm:
I don't have any objection to this, but I'm also not sure I have the right context to make the decision. I'll approve to unblock, and trust you to request reviews from anyone who should have a chance to review (maybe @MichaReiser ?)

MichaReiser:
Hmm. I don't have much context myself but I'll try to find the PR that added support for it in Ruff to understand what the motivation was back then.

MichaReiser:
I couldn't find a reason why PPC64 support was added in Ruff. Given that it's very easy for us to restore PPC64 support if users ask for it, I suggest we go ahead and merge this. This will at least give us signal on why we have to support PPC64 going forward (and for how long).

konstin:
The PR that added it is https://github.com/astral-sh/ruff/pull/45. My understanding is that we added it since it was supported by manylinux, where manylinux added it because it was supported by CentOS 7. The next version, CentOS 8, dropped support for big endian PPC64, and now even LLVM has dropped support (https://discourse.llvm.org/t/end-of-support-for-64-bit-big-endian-linux-on-powerpc/77164). PPC64 continuous to exist in one manylinux version, though that may be misleading in the sense that it's not getting removed from old version, but isn't supported by manylinux anymore? I opened a DPO thread to get some broader discussion on target support: https://discuss.python.org/t/what-platforms-should-wheels-be-provided-for-by-default/105822

Issue #2361, opened 2026-01-06
Ability to set output format with an environment variable


robertodr:
It would be useful to set the output format with an environment variable, _i.e._ `TY_OUTPUT_FORMAT`. Ruff has a similar environment variable.

MichaReiser:
Can you say a bit more about your use case?

robertodr:
I'm creating a custom pre-commit hook and would like to be able to switch the output format based on whether it's run in a github workflow or locally.

MichaReiser:
I see, thanks. Seems reasonable to me. Would you be interested in contributing the change? 

All that should be required is to set `env` similar to how it's done for the configuration: 

https://github.com/astral-sh/ruff/blob/e869046dcf75bd1581f6230e13b7bf3dec9b7353/crates/ty/src/args.rs#L125-L126

and then add a new test here

https://github.com/astral-sh/ruff/blob/e869046dcf75bd1581f6230e13b7bf3dec9b7353/crates/ty/tests/cli/main.rs#L96-L115

robertodr:
I can try!

Issue #1733, opened 2025-12-02
ty doesnt respect UV_PROJECT_ENVIRONMENT or `uv run`, leading it to incorrectly claim `unresolved-import`


brycedrennan:
In a docker image, when `UV_PROJECT_ENVIRONMENT=/usr/local`, `ty` cannot resolve third-party imports even though `uv run python` can import them just fine.

The use case is having uv install into vanilla python docker images and then running the type checker against code in that image.

I see in previous discussion that we dont want to complicate `ty` with all the logic of finding the right python installation, but I would have expected `uv run` to have solved that such that `uv run ty` "just works"

Also note that `ty` is not respecting the environment that it's installed into. This is contrary to a [recent comment:](https://github.com/astral-sh/ty/issues/684#issuecomment-3533238411) 
> We now respect the Python environment that ty itself is installed in

**Repro**
```bash
docker build -t ty-bug-repro .

# ty claims unresolved import:
docker run --rm ty-bug-repro uv run ty check
# error[unresolved-import]: Cannot resolve imported module `requests`

docker run --rm ty-bug-repro ty check
# error[unresolved-import]: Cannot resolve imported module `requests`

# But Python import works:
docker run --rm ty-bug-repro uv run main.py
# /usr/local/lib/python3.11/site-packages/requests/__init__.py
```

**Dockerfile:**
```dockerfile
FROM python:3.11-slim
COPY --from=ghcr.io/astral-sh/uv:latest /uv /bin/
ENV UV_PROJECT_ENVIRONMENT=/usr/local
WORKDIR /app
COPY . .
RUN uv sync 
```

**pyproject.toml:**
```toml
[project]
name = "ty-bug-repro"
version = "0.1.0"
requires-python = ">=3.11"
dependencies = ["requests", "ty"]
```

**main.py:**
```python
import requests
print(requests.__file__)
```
 

**Workarounds**
I don't love any of these. Mostly because every other command in my makefile is identical for local dev vs running in my docker image.

**1. Set PYTHONPATH:**
```dockerfile
ENV PYTHONPATH=/usr/local/lib/python3.11/site-packages
```

**2. Pass --python to ty:**
```bash
ty check --python /usr/local/bin/python3
```

**3. Fake a venv:**
```dockerfile
ENV VIRTUAL_ENV=/usr/local
RUN printf "home = /usr/local/bin\nversion = 3.11\n" > /usr/local/pyvenv.cfg
```

**4. Suppress the error**
```toml
[...]

Issue #269, opened 2025-05-08
`pre-commit` hook for `ty`


janosh:
looking forward to the https://github.com/astral-sh/ruff-pre-commit equivalent for `ty`

i understand `ty` is pre-release but i think there would be value in offering `pre-commit` integration even before the beta

lemorage:
Yes, I hope that as well. I use mypy a lot for static type checking, but the only available pre-commit hook is [mirrors-mypy](https://github.com/pre-commit/mirrors-mypy), which is maintained by the pre-commit team, not mypy itself.

AlexWaygood:
We don't plan on adding this right now because ty is still at a very early, experimental stage and we don't want to give the impression that it's ready for serious production use yet. When we're closer to feature-complete, however, we'll absolutely be adding this.

janosh:
@AlexWaygood makes sense.

for others wanting to ride out the experimental phase (given it might be a better UX than `mypy` 😄), you can use a `local` hook:

```yml
# .pre-commit-config.yaml
ci:
  skip: [ty]

repos:
  - repo: local
    hooks:
      - id: ty
        name: ty check
        entry: ty check .
        language: python
```

this will not work (e.g. in CI) unless you run `curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/ty/releases/download/0.0.0-alpha.7/ty-installer.sh | sh` first

the only issue i've had so far is i had to add this to `pyproject.toml`:


```toml
[tool.ty.rules]
unresolved-import = "ignore"
```

alon710:
> - repo: local
>     hooks:
>       - id: ty
>         name: ty check
>         entry: ty check .
>         language: python

You can also add the ignore directly to the pre commit config file

```yaml
repos:
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.11.0
    hooks:
      - id: ruff
        args: [--fix]
        files: \.py$
        exclude: ^(tests).*$ 
      - id: ruff-format
        files: \.py$
  - repo: local
    hooks:
      - id: ty
        name: ty check
        entry: uvx ty check . --ignore unresolved-import
        language: python
```

pelson:
[...]

PR #2626, opened 2026-01-26
Add `pre-commit` hook for `ty` type checker

chrimaho:
## Summary

This pull request adds a new pre-commit hook for Python type checking using the `ty` tool. This will help ensure type safety in Python code by running fast static analysis before commits. 

## Changes

- Introduce a new `pre-commit` hook configuration for the `ty` type checker.
- Added a new pre-commit hook for the `ty` type checker in `.pre-commit-hooks.yaml` to enable fast Python type checking.
- Define the hook with an entry point `ty check`.
- Specify the language as `python` and enable filename passing.
- Support Python file types for the hook.

## Issues

- #269
- #1733
- #2361

## References

I have followed the instructions given here:
https://pre-commit.com/#creating-new-hooks

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,8 @@
+- id: ty
+  name: ty type checker
+  description: An extremely fast Python type checker and language server, written in Rust.
+  entry: ty check
+  language: python
+  pass_filenames: true
+  types_or:
+    - python


AlexWaygood:
Thanks for the PR!

I think we agree that we want to provide a pre-commit hook for ty, but unfortunately I think we're still a bit undecided as a team about exactly what that should look like:
- Should we say that you have to repeat all your project dependencies in the `additional_dependencies` field, like mypy's pre-commit hook?
- Or should we try to detect a pre-existing `.venv` virtual environment and use the dependencies in that, _ignoring_ the isolated virtual environment that pre-commit sets up for us?
- Or should we have a pre-commit hook that uses both uv and ty to install your project dependencies into an entirely new isolated virtual environment, and then runs ty?

When we add a pre-commit hook for ty, we're also likely to have it in a separate repo (like `ruff-pre-commit`), so that the hook is able to add ty as a dependency. If the ty repository itself has a pre-commit hook listed, then unfortunately pre-commit will attempt to install ty from source, which will take a long time since ty is a Rust project. If there is a ty-pre-commit hook that depends on ty, however, installing that project from source will take no time at all, since it will be a minimal Python project that simply pulls a pre-compiled ty binary from PyPI as a Python dependency :-)

chrimaho:
Hi @AlexWaygood,
Makes sense. Thanks for the detailed input and insight. I appreciate it.
I love using `uv` and `ty`, and I look forward this solution once it is available!
Cheers

PR #2701, opened 2026-02-02
Fix empty body rule rendering

stefanvanburen:
Without the extra newline here, the bullets don't become a list and instead are mashed into the previous paragraph.

Ref: https://docs.astral.sh/ty/reference/rules/#empty-body


stefanvanburen:
Ah, I see this should go upstream to ruff instead: https://github.com/astral-sh/ruff/blob/8d78aa64174e79cf31fa564a80a466c18896202d/crates/ty/docs/rules.md?plain=1#L512-L513

stefanvanburen:
https://github.com/astral-sh/ruff/pull/23039

PR #2762, opened 2026-02-09
docs: Add command configuration for ty in neovim

joelostblom:
## Summary

For me, `ty` does not work in neovim unless I specify to run the `server` command in the config. Adding it to the docs might be helpful for first time users.

### DIFF
--- a/docs/editors.md
+++ b/docs/editors.md
@@ -48,7 +48,7 @@ For Neovim 0.11+ (with [`vim.lsp.config`](<https://neovim.io/doc/user/lsp.html#v
 
 ```lua
 vim.lsp.config('ty', {
-  -- Required: Start the ty LSP server
+  -- Required: Start the ty language server
   cmd = { "ty", "server" },
   -- Optional: Only required if you need to update the language server settings
   settings = {


BurntSushi:
Hmmm this shouldn't be necessary if you're [using `nvim-lspconfig`, which specifies `cmd` for you](https://github.com/neovim/nvim-lspconfig/blob/66fd02ad1c7ea31616d3ca678fa04e6d0b360824/lsp/ty.lua#L11).

joelostblom:
Yes, but the docs only mention to use `nvim-lspconfig` with neovim 0.10. For 0.11+, the native `vim.lsp.config` is mentioned, which does not contain a `ty` configuration natively.

Instead of adding a lspconfig recommendation also for 0.11+, I think it would be helpful to show the required lsp command for those who don't want to use neovim's native lsp capabilities without installing a separate plugin.

docs/editors.md
```diff
@@ -47,8 +47,10 @@ require('lspconfig').ty.setup({
 For Neovim 0.11+ (with [`vim.lsp.config`](<https://neovim.io/doc/user/lsp.html#vim.lsp.config()>)):
 
 ```lua
--- Optional: Only required if you need to update the language server settings
 vim.lsp.config('ty', {
+  -- Required: Start the ty LSP server
```
dhruvmanila:
```suggestion
  -- Required: Start the ty language server
```

BurntSushi:
I'm not sure we should be optimizing the docs for the case where `nvim-lspconfig` isn't being used. I'd rather the docs target `nvim-lspconfig` to keep things simpler (particularly if and when the ty config in`nvim-lspconfig` gets more complicated). For users looking to keep things lean and not use any plugins, I think they can go to `nvim-lspconfig` and copy the configuration from there.

I say this because [some of the configs can be rather complicated](https://github.com/neovim/nvim-lspconfig/blob/66fd02ad1c7ea31616d3ca678fa04e6d0b360824/lsp/rust_analyzer.lua). I don't know if ty's will ever get to that level, but if it does, I'd rather just point folks toward `nvim-lspconfig` instead of reproducing a complex config in our docs.

dhruvmanila:
> I'm not sure we should be optimizing the docs for the case where `nvim-lspconfig` isn't being used. I'd rather the docs target `nvim-lspconfig` to keep things simpler (particularly if and when the ty config in`nvim-lspconfig` gets more complicated). For users looking to keep things lean and not use any plugins, I think they can go to `nvim-lspconfig` and copy the configuration from there.
> 
> I say this because [some of the configs can be rather complicated](https://github.com/neovim/nvim-lspconfig/blob/66fd02ad1c7ea31616d3ca678fa04e6d0b360824/lsp/rust_analyzer.lua). I don't know if ty's will ever get to that level, but if it does, I'd rather just point folks toward `nvim-lspconfig` instead of reproducing a complex config in our docs.

That seems reasonable, I'll close this out.

joelostblom:
That makes sense! Maybe it's still possible to clarify the text a bit to highlight that nvim-lspconfig is recommended also for Neovim >=0.11? I made a suggestion in https://github.com/astral-sh/ty/pull/2785 as a continuation of this PR

PR #2772, opened 2026-02-10
Mention new goto definition behavior for class constructors

BurntSushi:
Ref https://github.com/astral-sh/ruff/pull/23071#issuecomment-3875176281

### DIFF
--- a/docs/features/language-server.md
+++ b/docs/features/language-server.md
@@ -37,7 +37,9 @@ see diagnostics for open files only, or for your entire workspace.
 ty powers several language server features that allow you to navigate a Python codebase:
 
 - **Go to Definition**: Jump to where a symbol is defined. ty resolves imports, function calls,
-    class references, and more.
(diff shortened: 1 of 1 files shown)


BurntSushi:
I'm not entirely convinced this is worth mentioning here since these docs seem like they are very high level, and this note is somewhat in the weeds.

dhruvmanila:
> I'm not entirely convinced this is worth mentioning here since these docs seem like they are very high level, and this note is somewhat in the weeds.

Yeah, I agree. I mainly wanted to point out about this being a subtle behavior and that we don't have a good place to put this in. I'm not sure if someone would notice this while reading this page, so I'd be inclined to avoid putting this in for now and opening a tracking issue related to documenting various behavior about the language server.

BurntSushi:
Closing in favor of https://github.com/astral-sh/ty/issues/2773

PR #2787, opened 2026-02-11
Add note about fallback behavior to `python` in `PATH`

BurntSushi:
Ref https://github.com/astral-sh/ruff/pull/22843#issuecomment-3884941254


BurntSushi:
Ref https://github.com/astral-sh/ruff/pull/23229

Issue #1976, opened 2025-12-17
LSP: Support callHierarchy/incomingCalls and callHierarchy/outgoingCalls


sr0lle:
According to the [docs](https://docs.astral.sh/ty/features/language-server/#feature-reference), neither of the [callHierarchy methods](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#callHierarchy_incomingCalls) is supported yet. Both methods are incredibly helpful when navigating or refactoring large projects, so it would be fantastic to have support for both and bring the LSP closer to being feature-complete.

mhuen:
I also ran into this. Currently using `pyright` for my applications but I needed something faster, which is how I ended up here. I do need support for the call hierarchy calls though, without which I cannot use `ty` for my use case. I don't usually code in Rust nor am I familiar with the ruff codebase, but perhaps the solution I have in astral-sh/ruff#25338 is helpful. That solution loosely follows call hierarchy conventions from `pyright` wherever it made sense. Known deviations are documented in the PR description

PR #2816, opened 2026-02-14
docs: add link for the call hierarchy tracking issue

toby-bro:
## Summary

This just adds a link in the documentation to the corresponding to the issue tracking call hierarchy feature (#1976 )

## Test Plan

The precommit hook passes, and when previewing the doc locally the change appears correct (and this is only a change to documentation).


carljm (approved):
Thank you!

PR #2852, opened 2026-02-19
Update the Python module (notably `find_ty_bin`) for parity with uv

zanieb:
Closes https://github.com/astral-sh/ruff/issues/18153

Copy of https://github.com/astral-sh/ruff/pull/23406


bengsparks:
This should also close https://github.com/astral-sh/ruff/issues/18153 because the ruff one has been merged as well, right? :)

PR #3017, opened 2026-03-11
docs: Update the Zed editor integration

lingyaochu:
## Summary

According to the PR zed-industries/zed#50662, the LSP adapters registry logic has been updated, and the corresponding documentation in Zed has changed as well.

This PR syncs the `ty` side to align with these changes.

## Test Plan

No test needed.

### DIFF
--- a/docs/editors.md
+++ b/docs/editors.md
@@ -77,9 +77,11 @@ You can enable ty and disable basedpyright by adding this to your `settings.json
   "languages": {
     "Python": {
       "language_servers": [
-        // Disable basedpyright and enable ty
+        // Enable ty and ruff,
+        // Other built-in servers (basedpyright, pyright, pylsp)
+        // are disabled by being omitted from this list.
         "ty",
-        "!basedpyright",
+        "ruff"
       ]
     }
   }


docs/editors.md
```diff
@@ -77,11 +77,9 @@ You can enable ty and disable basedpyright by adding this to your `settings.json
   "languages": {
     "Python": {
       "language_servers": [
-        // Disable basedpyright and enable ty, and otherwise
-        // use the default configuration.
+        // Disable basedpyright and enable ty
         "ty",
         "!basedpyright",
-        "..."
```
MichaReiser, line 84: Can you say more about why you removed `...`. Should we rephrase the comment instead of removing all other language servers (so that the documentation matches Zed's documentation in https://github.com/zed-industries/zed/pull/50662/changes)

lingyaochu: I removed the `"..."` because while its underlying logic (enabling all remaining registered adapters) hasn't changed, Zed's registry behavior has. Previously, adapters like `pylsp` and `pyright` and `ty` were lazy-loaded and didn't appear in the registry unless explicitly invoked. Because of this, a configuration like `["ty", "!basedpyright", "..."]` would coincidentally only enable `ruff`, making the old documentation’s claim that `...` meant "keep defaults" appear correct in practice.

However, after the registry logic change in zed-industries/zed#50662, all five Python adapters are now registered upfront. This means keeping the ellipsis would now result in an effective configuration of `["ty", "!basedpyright", "ruff", "pylsp", "pyright"]`, spawning redundant language servers that users don't actually want. While Zed’s own documentation might retain the `"..."` to explain the feature's mechanics, I believe our documentation for `ty` should prioritize providing a clean, "copy-pasteable" configuration. By removing the ellipsis, we ensure that users get exactly what they expect—`ty` enabled and `basedpyright` disabled—without any unintended side effects or performance overhead from extra LSPs.

lingyaochu: In case I wasn't clear, here's an [example](https://zed.dev/docs/configuring-languages#examples) from the main Zed docs that should give you a better idea of what I mean about the `"..."`

injust:
Oh nice, I was just about the make a PR for this change. See https://github.com/zed-industries/zed/issues/51416 for more context.

I wonder if ty's example config should also include `"ruff"`, since users are probably interested in using it along with ty. If you just remove the `"..."`, then Ruff would stop working.

docs/editors.md
```diff
@@ -77,11 +77,9 @@ You can enable ty and disable basedpyright by adding this to your `settings.json
   "languages": {
     "Python": {
       "language_servers": [
-        // Disable basedpyright and enable ty, and otherwise
-        // use the default configuration.
+        // Disable basedpyright and enable ty
         "ty",
         "!basedpyright",
```
injust: If removing `"..."`, then also remove `!basedpyright` entirely. There's really no point in leaving it.

docs/editors.md
injust: > I removed the `"..."` because while its underlying logic (enabling all remaining registered adapters) hasn't changed, Zed's registry behavior has. Previously, adapters like `pylsp` and `pyright` and `ty` were lazy-loaded and didn't appear in the registry unless explicitly invoked. Because of this, a configuration like `["ty", "!basedpyright", "..."]` would coincidentally only enable `ruff`, making the old documentation’s claim that `...` meant "keep defaults" appear correct in practice.
> 
> However, after the registry logic change in [zed-industries/zed#50662](https://github.com/zed-industries/zed/pull/50662), all five Python adapters are now registered upfront. This means keeping the ellipsis would now result in an effective configuration of `["ty", "!basedpyright", "ruff", "pylsp", "pyright"]`, spawning redundant language servers that users don't actually want. While Zed’s own documentation might retain the `"..."` to explain the feature's mechanics, I believe our documentation for `ty` should prioritize providing a clean, "copy-pasteable" configuration. By removing the ellipsis, we ensure that users get exactly what they expect—`ty` enabled and `basedpyright` disabled—without any unintended side effects or performance overhead from extra LSPs.

Thank you for explaining this! Finally I understand why the behaviour of `"..."` changed recently.

docs/editors.md
lingyaochu: You're absolutely right. I had a misunderstanding of how the configuration merges: I previously thought that servers not explicitly mentioned would still follow the default settings, which is why I didn't include `"ruff"` and thought I needed `"!basedpyright"`.

Since I actually want to keep `"ruff"` enabled alongside `"ty"`, I should explicitly list both and remove the unnecessary `"!basedpyright"` entry. I'll update the PR after some manual testing. Thanks for pointing this out!

docs/editors.md
injust: > Can you say more about why you removed `...`. Should we rephrase the comment instead of removing all other language servers (so that the documentation matches Zed's documentation in https://github.com/zed-industries/zed/pull/50662/changes)

Note that https://github.com/zed-industries/zed/pull/51427/changes was just merged, which makes Zed's docs on `"..."` more clear that it _enables_ all other languages servers.

PR #3073, opened 2026-03-18
Typing FAQ: New entry explaining invariance

sharkdp:
## Summary

Now that we do not union in `Unknown` into element types of invariant collections anymore, this question comes up more frequently. I'm taking a first stab at trying to explain this in a concise way. I am also planning to add a diagnostic hint that would be emitted in those situations, which could refer to this FAQ entry.

## Rendered

image

### DIFF
--- a/docs/reference/typing-faq.md
+++ b/docs/reference/typing-faq.md
@@ -94,19 +94,48 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
 
 ## Why can't I use `list[Subtype]` when a `list[Supertype]` is expected? { #invariant-generics }
 
-Let's say you have a class hierarchy with an `Entry` base class and a `Directory` subclass. Since
+Let's say you have a class hierarchy with an `Entry` base class as well as `Directory` and `File` subclasses. Since
 a `Directory` *is* an `Entry`, you can use it everywhere an `Entry` is expected.
 You might therefore expect a `list[Directory]` to
 be usable in any context where a `list[Entry]` is expected, but this is not
 the case. The reason for this is mutability:
 
 ```py
+# Setup of `Entry`, `Directory`, and `File` classes (1)
+
 def modify(entries: list[Entry]):
-    entries.append(File("README.txt"))
+    entries.append(File("README.txt")) # mutation
 
 directories: list[Directory] = [Directory("Downloads"), Directory("Documents")]
 modify(directories)  # ty emits an error on this call
 ```
+
+1. The full example might look like this:
+
+    ```py
+    from dataclasses import dataclass
+
+    @dataclass
+    class Entry:
+         path: str
+         def size_bytes(self) -> int: ...
+
+    @dataclass
+    class Directory(Entry):
+        def children(self) -> list[Entry]: ...
+
+    @dataclass
+    class File(Entry):
+        def content(self) -> bytes: ...
+
+    def modify(entries: list[Entry]):
+        entries.append(File("README.txt")) # mutation
+
+    directories: list[Directory] = [Directory("Downloads"), Directory("Documents")]
+    modify(directories)  # ty emits an error on this call
+    ```
+
+    You can try it out in [this playground example](https://play.ty.dev/01013e73-da54-40c4-a9c5-2af269abda9d).
 
 The `modify` call mutates the contents of the `directories` list. After this call,
 it contains two directories *and one `File`*, which clearly violates the


docs/reference/typing-faq.md
```diff
@@ -92,6 +92,60 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
 
     If you need this for `complex`, you can use `ty_extensions.JustComplex` in a similar way.
 
+## Why is this `list`/`set`/`dict` not assignable to this other collection type? What is invariance?
+
+In Python, `bool` is a subtype of `int`: you can use `True` and `False` in any
```
sharkdp: `bool <: int` isn't the perfect subtyping example. Other languages do not consider `bool` to be a subtype of `int` and it causes some issues. But more importantly for this example, `bool` doesn't really add any new properties or methods that could be used to clearly demonstrate unsoundness.

On the other hand, I didn't really want to add a subclass hierarchy like `Employee <: Person` here for reasons of conciseness. I'm happy to reconsider this choice though.

docs/reference/typing-faq.md
```diff
@@ -92,6 +92,60 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
...
+instead. This type describes read-only sequences (that contain values of type
+`int`). `Sequence` is therefore covariant in its type parameter.
+
+If you cannot adapt the signature of the function you are calling, you can also
+widen the type of the argument by annotating `answers` with `list[int]`.
+
+!!! note
+
+    In a similar way, you can use [`Mapping[str, V]`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping)
+    as a covariant alternative to `dict[str, V]`.
```
sharkdp: I could go into more details here and explain that `Mapping` is only covariant in its value type, but still invariant in its key type. But that seemed like too much information. I could also remove this `note` entirely.

docs/reference/typing-faq.md
sharkdp: I could also try to rewrite the example with subtyping relationship like `int <: int | None`, which also wouldn't require introducing a whole subclass hierarchy.

sharkdp: > I'm happy to reconsider this choice though.

Just discussed with Alex that a custom class hierarchy might be better after all.

docs/reference/typing-faq.md
```diff
@@ -92,6 +92,60 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
 
     If you need this for `complex`, you can use `ty_extensions.JustComplex` in a similar way.
 
+## Why is this `list`/`set`/`dict` not assignable to this other collection type? What is invariance?
```
AlexWaygood: This heading uses two pieces of jargon that I think most Python developers new to typing -- and even many that are familiar with typing -- might not be familiar with: "assignability" and "invariance". To make this more accessible, what about using something like


```suggestion
## Why does ty complain when I try to use `list[Subclass]` when `list[Superclass]` was expected?
```

or


```suggestion
## Why does ty complain when I try to pass a `list[Subclass]` instance to a parameter expecting `list[Superclass]`?
```

docs/reference/typing-faq.md
```diff
@@ -92,6 +92,60 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
...
+
+In Python, `bool` is a subtype of `int`: you can use `True` and `False` in any
+context where an `int` is expected. You might therefore expect a `list[bool]` to
+be usable in any context where a `list[int]` is expected. However, this is not
+the case. The reason for this is (interior) mutability:
```
AlexWaygood: "interior mutability" is not a term I ever encountered until I started learning my first low-level programming language (Rust). I think I'd just stick with "mutability" here


```suggestion
the case. The reason for this is mutability:
```

docs/reference/typing-faq.md
```diff
@@ -92,6 +92,60 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
...
+
+!!! info
+
+    In type system terminology, we say `list` is *invariant*, which means that
+    a subtyping relationship between types `A` and `B` does not imply any
+    relationship between `list[A]` and `list[B]`. The same is true for other
+    collection types such as `set` or `dict`. In contrast, read-only collections
+    like `tuple` or `frozenset` are *covariant* in their type parameter. It
+    is safe to assign a `frozenset[bool]` to a `frozenset[int]` because the
+    contents can not be mutated.
```
AlexWaygood:
```suggestion
    In type system terminology, we say `list` is *invariant*, which means that
    just because `A` is a subtype of `B` does not mean that `list[A]` will be a
    subtype of `list[B]`. The same is true for other builtin collections such as
    `set` or `dict`. In contrast, read-only collections like `tuple` or
    `frozenset` are *covariant* in their type parameter. It is safe to assign a
    `frozenset[bool]` to a `frozenset[int]` because the contents cannot be
    mutated.
```

docs/reference/typing-faq.md
```diff
@@ -92,6 +92,60 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
...
+def show_list_and_sum(xs: list[int]):
+    print(", ".join(str(x) for x in xs) + f" (sum = {sum(xs)})")
+
+answers = [True, False, False, True]  # inferred as list[bool]
+show_list_and_sum(answers)  # should be fine because no mutation occurs
+```
+
+To prevent this, you can adapt the signature of `show_list_and_sum` to take an
+argument of type
+[`Sequence[int]`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence)
```
AlexWaygood: the reference docs for the `collections.abc` interfaces are, unfortunately, terrible, such that I wonder if it would be better to link to the [summary table](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes) at the top of the docs, or the documentation [here](https://docs.python.org/3/library/stdtypes.html#sequence-types-list-tuple-range) that describes more broadly what it "means" to be a sequence, which references that `collections.abc.Sequence` provides a formal definition of the interface.

I myself always just reference the [typeshed definition](https://github.com/python/typeshed/blob/9c855f37cbb81c3a7e0a2db03375b5b6bd643111/stdlib/typing.pyi#L658-L670), which is the source of truth for type checkers

docs/reference/typing-faq.md
AlexWaygood: I think I'd avoid adding more details here about the invariance of the `Mapping` value type, because it breaks the (usually pretty good) heuristic of "if it's immutable, it can be covariant", and nobody (including me) can remember off the top of their head why it's necessary for soundness for the value type to be invariant.

Keeping the note seems useful, though, because there usually isn't a better option than using `Mapping` unless you want to define a custom protocol. You can use things like `Container` or `Collection`, but these are much broader interfaces that don't really express what it means for something to be a [mapping](https://docs.python.org/3/glossary.html#term-mapping)

sharkdp: > why it's necessary for soundness for the value type to be invariant.

you mean.. the key type?

docs/reference/typing-faq.md
sharkdp: I added this in parentheses because some developers also think that `x = 0; x = 1` is a form of mutation. I think it doesn't hurt to use the term? It seems pretty self-explanatory ("something *inside* this collection is being mutated").

I'm also fine removing it though

docs/reference/typing-faq.md
sharkdp: > I wonder if it would be better to link to the [summary table](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes) at the top of the docs

I like that, thanks!

docs/reference/typing-faq.md
AlexWaygood: er... yes :)

docs/reference/typing-faq.md
sharkdp: I went with "Why can't I use `list[Subtype]` when a `list[Supertype]` is expected?" for now. I don't want to restrict it to subclass/superclass because it often comes up with union types as the supertype.

docs/reference/typing-faq.md
AlexWaygood: hmm, I don't personally find it self-explanatory. I think Python developers might be more familiar with the concept of "in-place" mutation, if that's the distinction we're trying to draw here?

sharkdp: ok, removed

docs/reference/typing-faq.md
```diff
@@ -92,6 +92,72 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
...
+
+Let's say you have a class hierarchy with an `Entry` base class and a `Directory` subclass. Since
+a `Directory` *is* an `Entry`, you can use it everywhere an `Entry` is expected.
+You might therefore expect a `list[Directory]` to
+be usable in any context where a `list[Entry]` is expected, but this is not
+the case. The reason for this is (interior) mutability:
+
+```py
+def modify(entries: list[Entry]):
+    entries.append(File("README.txt"))
```
carljm: I _think_ it's sufficiently self-evident that `File` is a distinct subclass of `Entry`, but that's entirely implicit and caught me out the first time I read through this -- this is the first time we see `File` mentioned.

It would be nice if our example only required two types, not three, but I'm not sure if there's a good realistic example with that shape. `X` vs `X | None` is one realistic case, but then it requires also understanding unions and subtyping.

docs/reference/typing-faq.md
```diff
@@ -92,6 +92,72 @@ A similar rule applies to `complex`, which is treated as `int | float | complex`
...
+be usable in any context where a `list[Entry]` is expected, but this is not
+the case. The reason for this is mutability:
+
+```py
+def modify(entries: list[Entry]):
+    entries.append(File("README.txt"))
+
+directories: list[Directory] = [Directory("Downloads"), Directory("Documents")]
+modify(directories)  # ty emits an error on this call
+```
```
AlexWaygood, line 111: I know that it makes the example frustratingly verbose, but I do think being able to copy and paste examples directly into a Python REPL and/or type checker playgrounds is quite valuable. So I would be inclined to defined these classes in the Python snippet:


````suggestion
```py
from dataclasses import dataclass

@dataclass
class Entry:
    path: str
    
class Directory(Entry):
    pass

class File(Entry):
    pass

def modify(entries: list[Entry]):
    entries.append(File("README.txt"))

directories: list[Directory] = [Directory("Downloads"), Directory("Documents")]
modify(directories)  # ty emits an error on this call
```
````

File: docs/reference/typing-faq.md (hunk 7 above)
Comment (sharkdp) [reply docs/reference/typing-faq.md]: > It would be nice if our example only required two types, not three

That's exactly what I said to Alex. The drawback of the class hierarchy example is that it requires three types if you really want to demonstrate a bug.

We did consider using `int | None`, but also thought that explaining subtyping of unions would make it even more complicated.

> I _think_ it's sufficiently self-evident that `File` is a distinct subclass of `Entry`, but that's entirely implicit and caught me out the first time I read through this -- this is the first time we see `File` mentioned.

I'll think about this

File: docs/reference/typing-faq.md (hunk 8 above)
Comment (sharkdp) [reply docs/reference/typing-faq.md:111]: I thought about adding a playground link somewhere...

Comment (AlexWaygood) [reply docs/reference/typing-faq.md:111]: if you want it more concise, you could go for more "stub-like" formatting:


````suggestion
```py
from dataclasses import dataclass

@dataclass
class Entry:
    path: str
    
class Directory(Entry): ...
class File(Entry): ...

def modify(entries: list[Entry]):
    entries.append(File("README.txt"))

directories: list[Directory] = [Directory("Downloads"), Directory("Documents")]
modify(directories)  # ty emits an error on this call
```
````

carljm: This would also address my comment about `File` coming from nowhere.

sharkdp: I now (ab)used inline annotations to hide the setup code away in a tooltip

image

PR #3121, opened 2026-03-24
Update documentation to reflect `type:ignore` changes

MichaReiser:
## Summary

Update the documentation to reflect the changes introduced in https://github.com/astral-sh/ruff/pull/24096

### DIFF
--- a/docs/suppression.md
+++ b/docs/suppression.md
@@ -42,6 +42,15 @@ To suppress multiple violations on a single line, enumerate each rule separated 
 sum_three_numbers("one", 5)  # ty: ignore[missing-argument, invalid-argument-type]
 ```
 
+To suppress specific rules for an entire file, place a `# ty: ignore[<rule>]` comment on its own
+line before any Python code:
+
+```python
+# ty: ignore[invalid-argument-type]
+
+sum_three_numbers(3, 2, "1")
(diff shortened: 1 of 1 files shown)


docs/suppression.md
```diff
@@ -42,6 +42,15 @@ To suppress multiple violations on a single line, enumerate each rule separated
 sum_three_numbers("one", 5)  # ty: ignore[missing-argument, invalid-argument-type]
 ```
 
+To suppress specific rules for an entire file, place a `# ty: ignore[<rule>]` comment on its own
+line before any Python code:
+
+```python
+# ty: ignore[invalid-argument-type]
```
charliermarsh, line 49: Is it critical here that there's a blank line between this and any Python code? Like is this different than:
```python
# ty: ignore[invalid-argument-type]
sum_three_numbers(3, 2, "1")
```

carljm: No, as long as there is nothing before it, it's a file-wide suppression comment either way. And we don't support previous-line suppression comments.

I suspect this formatting is just to emphasize that it's not a previous-line suppression comment. It's a bit unfortunate how easily it could look like one, though in real code it's probably not an issue, since there will likely be new lines, an import block, etc.

PR #3145, opened 2026-03-26
Bump version to 0.0.26

sharkdp:


CHANGELOG.md
```diff
@@ -1,5 +1,37 @@
 # Changelog
 
+## 0.0.26
+
+Released on 2026-03-26.
+
+### Bug fixes
+
+- Prevent `too-many-cycle` panics in more situations ([#24061](https://github.com/astral-sh/ruff/pull/24061))
+- Fixed false positive errors when unpacking a union of tuples into a function with optional parameters ([#23298](https://github.com/astral-sh/ruff/pull/23298))
```
sharkdp: This single entry is agent-generated. We could also remove it or someone provides a correction.

CHANGELOG.md
```diff
@@ -1,5 +1,37 @@
 # Changelog
 
+## 0.0.26
+
+Released on 2026-03-26.
+
+### Bug fixes
+
+- Prevent `too-many-cycle` panics in more situations ([#24061](https://github.com/astral-sh/ruff/pull/24061))
```
carljm:
```suggestion
- Prevent "too many cycle iteration" panics in more situations ([#24061](https://github.com/astral-sh/ruff/pull/24061))
```

(`too-many-cycle` is not an error code, it's just a description, so doesn't belong in backticks with dashes -- and it's missing "iterations")

CHANGELOG.md
carljm:
```suggestion
- Fixed false positives and false negatives when unpacking a union of tuples into a function call ([#23298](https://github.com/astral-sh/ruff/pull/23298))
```

CHANGELOG.md
```diff
@@ -1,5 +1,37 @@
...
+
+### Performance
+
+- Fix performance regression for narrowing on a larger `Literal` type([#24185](https://github.com/astral-sh/ruff/pull/24185))
+
+### Contributors
+
+- [@sharkdp](https://github.com/sharkdp)
+- [@Glyphack](https://github.com/Glyphack)
+- [@charliermarsh](https://github.com/charliermarsh)
```
carljm, line 31: I don't really care about it personally, but I kind of feel like this list ideally would include `Co-authored By` also -- I wrote most of #23298 :)

CHANGELOG.md
AlexWaygood: same suggestion but using the imperative mood:

```suggestion
- Fix false positives and false negatives when unpacking a union of tuples into a function call ([#23298](https://github.com/astral-sh/ruff/pull/23298))
```

CHANGELOG.md
AlexWaygood: sounds like you'll be vibecoding a PR to Rooster later today?

carljm: if only I cared enough

PR #3146, opened 2026-03-26
docs: use content tabs

LoicRiegel:
## Summary

Add content tabs in the documentation, showing both `pyproject.toml` and `ty.toml` variants. Same as in Ty's or Ruff’s documentation. Previously only `pyproject.toml` was shown, which might lead people to think that only this file is supported.

## Test Plan

Manual testing building the docs locally.

### DIFF
--- a/docs/exclusions.md
+++ b/docs/exclusions.md
@@ -4,21 +4,41 @@ ty automatically discovers all Python files in your project. You can customize w
 
 For example, with the following configuration, ty checks all Python files in the `src` and `tests` directories except those in the `src/generated` directory:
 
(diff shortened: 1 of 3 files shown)


AlexWaygood:
> Same as in Ty's or Ruff’s documentation.

this is ty's documentation -- did you mean "uv's or Ruff's documentation"? :-)

LoicRiegel:
> > Same as in Ty's or Ruff’s documentation.
> 
> this is ty's documentation -- did you mean "uv's or Ruff's documentation"? :-)

I meant ty, because those tabs were already used in a few places, just not everywhere :)

PR #3160, opened 2026-03-27
Fetch the cargo-dist binary directly instead of using the installer

zanieb:
See https://github.com/astral-sh/uv/pull/18731

### DIFF
--- a/dist-workspace.toml
+++ b/dist-workspace.toml
@@ -5,6 +5,7 @@ version = "0.0.26"
 
 # Config for 'dist'
 [dist]
+allow-dirty = ["ci"]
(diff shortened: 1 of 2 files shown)


charliermarsh (approved):

dist-workspace.toml
```diff
@@ -5,6 +5,7 @@ version = "0.0.26"
 
 # Config for 'dist'
 [dist]
+allow-dirty = ["ci"]
```
zanieb, line 8: We should upstream this change so we don't need to dirty the workflow for it.

In uv, we have other dirtied changes. It's not a big deal these days since cargo-dist is updated far less often.

PR #3292, opened 2026-04-17
Add version metadata publishing to the release process

zanieb:


.github/zizmor.yml
```diff
@@ -1,5 +1,6 @@
 rules:
   secrets-outside-env:
     ignore:
-      # TODO: move the ASTRAL_DOCS_PAT secret to the release environment
+      # TODO: move these secrets to the release environment
       - publish-docs.yml
+      - publish-versions.yml
```
zanieb, line 6: This is actually done, but will fix separately.

.github/workflows/publish-versions.yml
```diff
@@ -0,0 +1,76 @@
...
+            --label "automation"
+
+      - name: "Merge Pull Request"
+        if: ${{ !fromJson(inputs.plan).announcement_tag_is_implicit }}
+        working-directory: astral-versions
+        env:
+          GITHUB_TOKEN: ${{ secrets.ASTRAL_VERSIONS_PAT }}
+        run: |
+          # Wait for PR to be created before merging
+          sleep 10
```
zsol, line 75: ugh. I don't like this but I couldn't come up with a better suggestion in 2 minutes

zsol (approved):

.github/workflows/publish-versions.yml
zanieb: Unless I'm mistaken, it's copied from all the other projects. I think the "fix" is to hit the API looking for the pull in a loop, but why deal with that.

PR #3309, opened 2026-04-22
Add a 'Diagnostics improvements' section to the changelogs

AlexWaygood:
## Summary

ty PRs tagged with `diagnostics` are usually solely focussed on improving diagnostics. It makes sense for these to usually go in their own changelog section

## Test Plan

N/A


pyproject.toml
```diff
@@ -124,6 +124,7 @@ trim-title-prefixes = ["[ty]"]
 [tool.rooster.section-labels]
 "Breaking changes" = ["breaking"]
 "Preview features" = ["preview"]
+"Diagnostics improvements" = ["diagnostics"]
```
MichaReiser: I'm leaning towards removing `improvements`. It's sort of implicitly given that this is always our goal and we also omit it from `CLI` and `LSP Server`.

AlexWaygood:
```suggestion
"Diagnostics" = ["diagnostics"]
```

MichaReiser (approved):

PR #3335, opened 2026-04-25
update typing-features and faqs

### LINKED ISSUE #3266: Update the "gradual guarantee" section in the docs
The example in that section still works, but it might be worth checking if that section (and the title) imply a wider support for the gradual guarantee than what we provide, given some recent changes. At the very least, the "info" box should be updated.

https://docs.astral.sh/ty/features/type-system/#gradual-guarantee

carljm:
Closes #3266.

This updates our typing-features and typing-faq docs to account for recent changes in ty's behavior. Both the "gradual guarantee" and "fixpoint iteration" sections no longer describe our actual behavior. In both cases, I tried different ways to update the section, but ended up feeling like there was nothing that distinctive or interesting to say, so I just removed the sections entirely.

For gradual guarantee, I don't think we should have a section with that name, since in general we don't take that as a guiding principle anymore. The remaining singleton case is very niche, and already discussed in the FAQ.

For fixpoint iteration, we still do that of course, but for performance reasons we have since decided to promote literal unions much more quickly in all cyclic cases. So I wasn't able to construct any realistic case anymore where this "union of literals up to some limit" behavior is visible.

### DIFF
--- a/docs/features/type-system.md
+++ b/docs/features/type-system.md
@@ -169,55 +169,5 @@ def to_json(person: Person):
 
 (Full example in the [playground](https://play.ty.dev/34a227bb-93d5-405e-86c3-72f57ec5642e))
 
-## Gradual guarantee
-
-ty generally tries to avoid emitting false positive type errors in untyped code. The following
-snippet does not produce any type errors when checked with ty (whereas other type checkers make the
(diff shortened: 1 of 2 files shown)


sharkdp (approved):
Thanks!

PR #3368, opened 2026-04-28
Bump version to 0.0.33

sharkdp:


CHANGELOG.md
```diff
@@ -1,5 +1,78 @@
...
+
+- Add support for go-to definition in literal enum member inlay hints ([#24792](https://github.com/astral-sh/ruff/pull/24792))
+- Add support for "baking" keyword argument inlay hints into the source code ([#24667](https://github.com/astral-sh/ruff/pull/24667))
+- Don't allow inlay hint edits when introducing a non global scope symbol ([#24797](https://github.com/astral-sh/ruff/pull/24797))
+- Omit semantic highlighting for unresolved symbols ([#24718](https://github.com/astral-sh/ruff/pull/24718))
+
+### Core type checking
+
+- Support narrowing with aliased conditional expressions ([#24302](https://github.com/astral-sh/ruff/pull/24302))
+- Prefer declared type in more situations (if inferred and declared types are mutually assignable) ([#24802](https://github.com/astral-sh/ruff/pull/24802))
```
MichaReiser: Should we call this out more explicitly? It's a fairly big change and it may remove the need for `cast` in many places

MichaReiser (approved):

CHANGELOG.md
sharkdp: Sure. Are you looking for a different wording or a more prominent position? I think I would personally just release this, and then add a comment on all of the linked tickets that this is now supported (and that users may be able to remove `cast`s)?

MichaReiser: I consider it a breaking change, which is why I would add some prose explanation to the top. Similar to what you'd share in our weekly update

PR #3373, opened 2026-04-28
Release: move 'diagnostics' section further down

sharkdp:
## Summary

I was surprised to see the "Diagnostics" section above "Bug fixes" when I did the release. The ordering of these entries seems to control the relative placement of sections in the CHANGELOG, so I'm proposing to move the "Diagnostics" section further down.


AlexWaygood (approved):
oops, I didn't realise the order was load-bearing there

PR #3402, opened 2026-05-03
Build riscv64 manylinux binary

Xeonacid:
## Summary

Basically same as https://github.com/astral-sh/ruff/pull/19819

## Test Plan

Let CI test itself :P


.github/workflows/build-binaries.yml
MichaReiser: Do we need to add `libatomic1` here, similar to what was done in the Ruff PR?

Xeonacid: Oops, I forgot to push this. It's needed yeah. Thanks for the review!

Xeonacid:
I guess the prek CI failure is unrelated to this PR 👀

MichaReiser:
Thank you

PR #3513, opened 2026-05-22
Document `completeFunctionParentheses` editor setting

MichaReiser:
Documents the editor setting introduced by [astral-sh/ruff#25305](https://github.com/astral-sh/ruff/pull/25305).

### DIFF
--- a/docs/reference/editor-settings.md
+++ b/docs/reference/editor-settings.md
@@ -510,6 +510,67 @@ include symbols not currently in scope but available in your environment.
     }
     ```
 
+### `completeFunctionParentheses`
+
+Whether accepting a function, method, or class completion also inserts parentheses and places the
+cursor inside them.
+
+**Default value**: `false`
+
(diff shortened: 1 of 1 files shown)


docs/reference/editor-settings.md
```diff
@@ -510,6 +510,67 @@ include symbols not currently in scope but available in your environment.
     }
     ```
 
+### `completeFunctionParentheses`
```
lerebear, line 513: Should we also document that the client/editor must support LSP completion item snippets in order to use this feature? Since we silently disable the feature when the client doesn't support snippets, I think documenting as much might help explain otherwise surprising behaviour.

lerebear (approved):

docs/reference/editor-settings.md
MichaReiser: I followed your recommendation to always add the parentheses, even if the client does not support snippets.

PR #3523, opened 2026-05-23
scripts/update_schemastore: add text=True to git revision check_output calls

Dev-X25874:
## Summary

`main()` in `scripts/update_schemastore.py` calls `check_output` twice to read
git SHAs — once for the expected ruff submodule revision and once for the actual
HEAD of the submodule — but neither call passes `text=True`. Without it,
`check_output` returns `bytes`, and when those values are interpolated into the
mismatch warning f-string the user sees raw bytes repr in their terminal:

    The ruff submodule is at b'def456...' but main expects b'abc123...'

instead of the intended:

    The ruff submodule is at def456... but main expects abc123...

The same `main()` already uses `text=True` correctly for the `current_sha`
call on line 81. This PR makes the two revision-check calls consistent with
that pattern.

## Test Plan

Added `scripts/tests/test_update_schemastore.py` with a test that monkey-patches
`check_output` to return the mismatched-revision case, captures stdout, and
asserts the output contains plain SHA strings without any `b'...'` wrapper.
The test fails against the original code and passes after the fix.

    python -m pytest scripts/tests/test_update_schemastore.py -v


Dev-X25874:
The only remaining failure is a mdformat blank line issue caused by GitHub's web editor mangling the formatting. Happy for a maintainer to fix it directly or I can push a local fix if needed.

dhruvmanila (approved):
Thank you.

I'm going to remove the test file given that these are scripts and we usually don't write tests for it. I'm also updating the `print` call to use `!r` so that the quotes are present in the output.

Dev-X25874:
Thank you for the review and for taking the time to clean things up! Totally understand the no-tests policy for scripts — good to know for future contributions. The !r approach is a neat solution. Appreciate you merging this! 🙏

PR #3526, opened 2026-05-23
fix(jupyter): update ruff submodule with line_index safe clamp

### LINKED ISSUE #3525: panic: `index out of bounds` in `ruff_source_file/src/line_index.rs:269` while emitting `unused-ignore-comment` for a notebook cell
**Version**: `ty 0.0.39 (020512517 2026-05-22)`

## Summary

`ty check` panics with `index out of bounds` while emitting an `unused-ignore-comment` diagnostic for a `# ty: ignore[…]` (or PEP 484 `# type: ignore`) directive inside a Jupyter notebook code cell. The panic occurs in `ruff_source_file/src/line_index.rs:269` during span computation for the diagnostic. The panic was reproducible on at least two notebooks during a large refactor pass; I was unable to extract a single-cell minimal repro afterwards.

## Observed panics

Captured during a mass conversion of `# type: ignore[<ty-rule>]` → `# ty: ignore[<ty-rule>]` across a notebook corpus. In each case the immediate prior diagnostic was a `warning[unused-ignore-comment]` for a notebook cell suppression with the autofix `Remove the unused suppression comment` hint:

```
warning[unused-ignore-comment]: Unused blanket `ty: ignore` directive
 --> path/to/notebook-a.ipynb:cell 3:47:28
thread 'main' panicked at crates/ruff_source_file/src/line_index.rs:269:13:
index out of bounds: the len is 70 but the index is 74
```

```
warning[unused-ignore-comment]: Unused `ty: ignore` directive
 --> path/to/notebook-b.ipynb:cell 3:1:48
thread 'main' panicked at crates/ruff_source_file/src/line_index.rs:269:13:
index out of bounds: the len is 31 but the index is 63
```

## Observed conditions

Common across the panicking notebooks:

- Code cells whose `source` is stored as a JSON array of **single-character** strings (each element is one character of the cell's text), as opposed to the usual one-string-per-line representation. (Likely produced by some prior notebook tooling pass.)
- A `# ty: ignore[<rule>]` (or `# type: ignore`) directive on a line where the named rule does not match any diagnostic emitted on that line → ty marks it `unused-ignore-comment`.
- The panic occurs during emission of that warning's location/span, before the diagnostic is fully printed.

The PEP 484 form (`# type: ignore` bare or `# type: ignore[X]` no-op) also triggered the panic in one of the cases.

## What I tried (could not isolate)

- Single-line notebook with per-char source + unused `# ty: ignore[X]` → no panic
- Multi-line notebook with per-char source + unused `# ty: ignore[X]` mid-cell → no panic
- Two unused suppressions in a per-char-split cell → no panic
- Replaying the same edit on the original file outside the broader pass → no panic

The panic appears to need some state I haven't pinpointed. Happy to provide the affected notebook files and the edit script that reliably triggered the panic during the original session.

## Environment

- macOS arm64
- Python 3.11
- ty 0.0.39 (uv-managed)

iAbhi001:
This PR fixes a thread panic (index out of bounds) inside the ruff linter engine that occurs when running diagnostic or suppression checks against non-standard Jupyter Notebook layouts.

When a Jupyter notebook contains cell sources stored as fragmented, single-character string JSON arrays (instead of standard line-by-line arrays), the virtual text coordinate calculation can get out of sync. This results in tracking offsets that exceed the length of the actual string segment passed to the diagnostics subsystem. When LineIndex::source_location attempts to slice the text up to that character offset, Rust triggers a runtime bounds panic and crashes the execution thread.

To fix this, the incoming offset in source_location is now explicitly clamped to the maximum length of the text segment. This keeps the slicing boundaries safe, allowing the diagnostic framework to gracefully fall back to the end of the text segment instead of blowing up the process.

Fixes: #3525


MichaReiser:
Thank you. Would you mind opening this PR against https://github.com/astral-sh/ruff/

Issue #2902, opened 2026-02-24
Documentation: Editor integration: Emacs


FabriceSalvaire:
### Summary

I open this issue to draft an Emacs section in https://docs.astral.sh/ty/editors

# Support

- Ty works with Emacs [flymake](https://www.gnu.org/software/emacs/manual/html_node/flymake/index.html).
- Ty now works with [flycheck]( https://www.flycheck.org) and [eglot](https://www.gnu.org/software/emacs/manual/html_mono/eglot.html) with [flycheck support for eglot](https://github.com/flycheck/flycheck-eglot) (as of 2016/02/24).

# Setup

Add this snippet in `init.el`
```lisp
;; load flycheck, eglot, and flycheck-eglot packages
;;   or use Flymake package
;;   see package doc

(use-package flycheck
  :ensure t

  :init
  (global-flycheck-mode)
  ;; A minor mode to show Flycheck error messages in a popup.
  (flycheck-pos-tip-mode)
)

;; A minor-mode for Flycheck which colors the mode line according to the Flycheck state of the current buffer.
;;   https://github.com/flycheck/flycheck-color-mode-line
(use-package flycheck-color-mode-line
  :after (flycheck)
  :hook (flycheck-mode . flycheck-color-mode-line-mode)
  )

(use-package eglot
  :ensure t

  :commands eglot
  ;; auto start eglot for those modes
  :hook ((python-mode python-ts-mode) . eglot-ensure)
  :custom
  (eglot-autoshutdown t)
  )

;; Customize eglot-server-programs to use ty as lsp
(with-eval-after-load 'eglot
  (add-to-list 'eglot-server-programs
               `((python-mode python-ts-mode) . ("ty" "server"))))

;; Flycheck support for eglot
;;   https://github.com/intramurz/flycheck-eglot
(use-package flycheck-eglot
  :ensure t
  :after (flycheck eglot)
  :config
  (global-flycheck-eglot-mode 1)
  )
```

# Eglot Inlay Hints

Eglot Inlay Hints mode show implicit kwargs in the buffer.

Use `M-x eglot-inlay-hints-mode` **to toggle inlay hints mode** which are quite annoying (buggy) while editing code !!!

**inlay hints** are displayed using the three faces `eglot-type-hint-face`, `eglot-parameter-hint-face`, and `eglot-inlay-hint-face`, which all inherit from `eglot-inlay-hint-face`.

dhruvmanila:
[...]

PR #3528, opened 2026-05-24
docs: add Flycheck note to Emacs editor integration

jpatel3:
## Summary

Adds a short note to the Emacs section of the editor integration docs explaining how Flycheck users can route ty's diagnostics through the [flycheck-eglot](https://github.com/flycheck/flycheck-eglot) bridge.

The current Eglot snippet (added in #3107) covers the simple case, but the original request in #2902 specifically asked about Flycheck-based setups. This adds the minimal additional snippet for that path without changing the existing recommended setup. Modeled on the multi-subsection layout used by Ruff's [Emacs docs](https://docs.astral.sh/ruff/editors/setup/#emacs).

Refs #2902.

## Test Plan

- `mdformat docs/editors.md --check` passes
- `markdownlint-cli@0.48.0 docs/editors.md` passes
- Verified `flycheck/flycheck-eglot` is the official upstream package

### DIFF
--- a/docs/editors.md
+++ b/docs/editors.md
@@ -137,15 +137,7 @@ ty can be utilized as a language server via the built-in [Eglot](https://www.gnu
 
 If you prefer to view ty's diagnostics through [Flycheck](https://www.flycheck.org/), the
 [flycheck-eglot](https://github.com/flycheck/flycheck-eglot) package bridges Eglot's diagnostics
-to Flycheck:
-
-```elisp
-(use-package flycheck-eglot
-  :ensure t
-  :after (flycheck eglot)
-  :config
-  (global-flycheck-eglot-mode 1))
-```
+to Flycheck.
 
 ## Other editors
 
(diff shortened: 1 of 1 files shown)


docs/editors.md
```diff
@@ -135,6 +135,18 @@ ty can be utilized as a language server via the built-in [Eglot](https://www.gnu
...
+[flycheck-eglot](https://github.com/flycheck/flycheck-eglot) package bridges Eglot's diagnostics
+to Flycheck:
+
+```elisp
+(use-package flycheck-eglot
+  :ensure t
+  :after (flycheck eglot)
+  :config
+  (global-flycheck-eglot-mode 1))
+```
```
dhruvmanila: I think I'd prefer to remove the code snippet mainly because it's the same as the one in the `flycheck-eglot` configuration. This would make sure any future changes to the package doesn't require any changes in our documentation.

PR #3532, opened 2026-05-25
docs: set Eglot `:language-id` so ty works with `python-base-mode`

### LINKED ISSUE #2937: No diagnostics for LSP client that uses different language identifier
> @markhepburn Thank you for providing the logs, that's very useful and it does seem that yours might be a separate issue.
> 
> ```
> "languageId":"python-base"
> ```
> 
> Based on the logs for the open document request, it seems that the server uses `python-base` as the language ID while ty only recognizes `python`:
> 
> https://github.com/astral-sh/ruff/blob/ee277469d3bddb0d5fd59e308263059561be1e0a/crates/ty_server/src/document/text_document.rs#L39-L46
> 
> So, this might be an easy fix. 

 _Originally posted by @dhruvmanila in [#2824](https://github.com/astral-sh/ty/issues/2824#issuecomment-3983054627)_

But, the spec does recommend to use `python`: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentItem

jpatel3:
## Summary

- Closes #2937.
- Adds `:language-id "python"` to the Eglot `eglot-server-programs` entry in the Emacs editor-integration docs so the recommended snippet works out of the box.
- Eglot derives the LSP `languageId` from the major-mode symbol, so the previous `python-base-mode` entry caused Eglot to advertise documents as `python-base`. The ty server only accepts the `python` language id ([`text_document.rs`](https://github.com/astral-sh/ruff/blob/main/crates/ty_server/src/document/text_document.rs)), so users were silently getting no diagnostics.
- Adds a short prose note explaining the constraint, so readers who copy the snippet (or adapt it for a different mode) understand why the entry is shaped this way. Fix matches the one suggested upstream in [eglot#1580](https://github.com/joaotavora/eglot/discussions/1580).

## Test plan

- [x] Visually proof the rendered Markdown
- [x] Verified the snippet shape against the eglot discussion accepted answer

### DIFF
--- a/docs/editors.md
+++ b/docs/editors.md
@@ -130,10 +130,15 @@ ty can be utilized as a language server via the built-in [Eglot](https://www.gnu
 ```elisp
 (with-eval-after-load 'eglot
   (add-to-list 'eglot-server-programs
-               '(python-base-mode . ("ty" "server"))))
+               '((python-base-mode :language-id "python") . ("ty" "server"))))
 
 (add-hook 'python-base-mode-hook 'eglot-ensure)
 ```
+
+The `:language-id "python"` entry ensures Eglot advertises the document as
+`python` rather than `python-base` when registering `python-base-mode`. The ty
+server only recognises the `python` language identifier and will silently skip
+documents that announce themselves under any other identifier.
 
 If you prefer to view ty's diagnostics through [Flycheck](https://www.flycheck.org/), the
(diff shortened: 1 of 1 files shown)


dhruvmanila (approved):
Thank you! The explanation is useful but I think it's not really required in the setup documentation here, so I've removed it. Let me know if you think otherwise.

PR #3582, opened 2026-05-29
Document that call hierarchy is now supported

MichaReiser:


MichaReiser:
I think one entry for all the call hierarchy requests is sufficient

AlexWaygood (approved):

PR #3727, opened 2026-06-10
Automatically trigger new ty-pre-commit releases when we cut a release

AlexWaygood:
## Summary

This adds an additional step to our release workflow so that a new release auto-triggers a new ty-pre-commit release immediately. (Currently, the only trigger is a cron job over at ty-pre-commit that runs every four hours.) The new workflow works in exactly the same way as the existing workflow for Ruff.

In order for this to work, we will also need to add a `TY_PRE_COMMIT_PAT` secret, with Actions write access to `astral-sh/ty-pre-commit`, to the release environment. Ruff has the equivalent `RUFF_PRE_COMMIT_PAT`; without this secret, the new job will fail at dispatch time.

### DIFF
--- a/.github/workflows/notify-dependents.yml
+++ b/.github/workflows/notify-dependents.yml
@@ -1,0 +1,31 @@
+# Notify downstream repositories of a new release.
+#
+# Assumed to run as a subworkflow of .github/workflows/release.yml; specifically, as a post-announce
+# job within `cargo-dist`.
+name: "[ty] Notify dependents"
+
+on:
+  workflow_call:
+    inputs:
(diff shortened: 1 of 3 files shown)


MichaReiser (approved):

AlexWaygood:
> In order for this to work, we will also need to add a `TY_PRE_COMMIT_PAT` secret, with Actions write access to `astral-sh/ty-pre-commit`, to the release environment. Ruff has the equivalent `RUFF_PRE_COMMIT_PAT`; without this secret, the new job will fail at dispatch time.

I think this is now done

PR #3783, opened 2026-06-16
Expand mypy and Pyright rule mapping

AlexWaygood:
## Summary

- expand the migration table with additional mypy and Pyright correspondences
- point to Ruff rules when lint coverage exists outside ty
- link open ty and Ruff issues for checks that do not have an equivalent yet
- document that ty checks unannotated function bodies unconditionally
- add links to documentation for all mypy/ty/pyright rules

This aims to partially address some feedback we received that it was hard to know exactly which rules to enable if you wanted to approximate mypy's `--strict` mode.

### Table header and initial rows

Rendered mapping table header and initial rows

### Tracked gaps and additional mappings

Rendered tracked gaps and additional mappings

### Ruff alternatives and final rows

Rendered Ruff alternatives and final rows


AlexWaygood:
This makes the Markdown table here truly monstrous in its size and complexity. We could consider using an embedded TOML file or something, but I'd prefer to do that as a followup rather than in this PR, which is purely about content.

sharkdp (approved):
Thank you very much!

PR #3785, opened 2026-06-16
Document fixable rules

### LINKED ISSUE #3774: Document what codes the `--fix` flag fixes.
Since I couldn't find it in the documentation, I asked an agent to make a clone of the `ty` repo and figure out what issue codes the `--fix` flag actually fixes. It would be nice if this information were easier to look up.

<details>
<summary>
image of what the agent found
</summary>

Image

</details>

popsiclelmlm:
## Summary

Document which diagnostics currently have safe fixes applied by `ty check --fix`.

## Reproduction

Before this change, the CLI reference documented that `--fix` applies fixes, but the docs did not list which rule codes currently provide safe automatic fixes.

## Root cause

The public docs described the flag behavior but did not connect it to the specific fixable diagnostics implemented today.

## Changes

- Add a "Fixes" section to the rules guide.
- List `redundant-cast` and `unused-ignore-comment` as currently fixable by `ty check --fix`.
- Clarify that editor quick fixes and `ty check --add-ignore` are separate from diagnostic-specific safe fixes.

## Tests

- `npx prettier --prose-wrap always --check docs/rules.md`
- `/tmp/ty-docs-venv/bin/mkdocs build -f mkdocs.yml --strict`

## Screenshots/Logs

Not applicable; docs-only text change.

Fixes #3774

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -13,7 +13,9 @@ specific pattern and can be turned on or off depending on your project’s needs
 Each rule has a configurable level:
 
 - `error`: violations are reported as errors and ty exits with an exit code of 1 if there's any.
-- `warn`: violations are reported as warnings. Depending on your configuration, ty exits with an exit code of 0 if there are only warning violations (default) or 1 when using `--error-on-warning`.
+- `warn`: violations are reported as warnings. Depending on your configuration, ty exits with an
+  exit code of 0 if there are only warning violations (default) or 1 when using
+  `--error-on-warning`.
 - `ignore`: the rule is turned off
 
 You can configure the level for each rule on the command line using the `--warn`, `--error`, and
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Thanks. I feel like this could quickly go out of date, however -- I'd prefer it if we could find a way to generate this list somehow?

popsiclelmlm:
That makes sense. I traced the current implementation before opening this: `ty check --fix` applies diagnostics with `Applicability::Safe`, and the current safe fixes I found are `unused-ignore-comment` and `redundant-cast`. The problem is that this information is not represented in the rule metadata that generates the reference docs, so the table here would indeed be hand-maintained.

I’ll look into whether fixability can be exposed through the generated rule metadata instead. If that turns out to be larger than this docs PR, I’m fine closing this and leaving #3774 for a generated approach.

MichaReiser:
What we du in Ruff is that the fixability is part of the RuleMetadata

https://github.com/astral-sh/ruff/blob/b017fd1a55a7f29d5b6f08ef90b805a4f8731169/crates/ruff_linter/src/violation.rs#L14-L19

We can make `fixability` an optional argument to `declare_lint`. 

Now, there's also a risk that this metadata gets out of date. That's why, in tests, we verify that:

* For a rule with `Always`, error if a `Diagnostic` misses a fix
* For a rule with `None` (I think it should be called `Never`), error if a `Diagnostic` has a fix

For ty, I suggest we add a similar check to `ty_test`.

PR #3793, opened 2026-06-17
Add PGO to native ty release builds

MichaReiser:
Train Darwin and Linux GNU release wheels on the ecosystem analyzer corpus before the final maturin build, then feed the merged profile into the shipping build.
Add local helpers for reproducing PGO builds and corpus comparisons.
Testing: Ran focused hooks, a clean train-only PGO corpus run, and Build binaries CI.


MichaReiser:
Closing for now. The results ranged from +10 to -10%. Most projects improved, but a few also regressed

PR #3794, opened 2026-06-17
fix: Resolve attribute access on TypeVars bound to a type[...] form

### LINKED ISSUE #3782: Attribute access on a TypeVar bound to `type[...]` is unresolved (works for plain `type[X]` and instance-bound TypeVars)
### Summary

Accessing a class attribute on a value whose type is a `TypeVar` with a `type[...]`
upper bound produces a false `unresolved-attribute`, even though the attribute is
defined on the bound class(es). The same access works for:

- a plain `type[X]` parameter (no TypeVar), and
- a TypeVar bound to an *instance* union (`A | B`) — fixed in 0.0.43 (ruff #25561).

So the gap is specifically TypeVars whose **upper bound is a `type[...]` form**
(both the single `type[A]` and the union `type[A] | type[B]` cases fail).

### Reproduction

```python
from typing import ClassVar, TypeVar

class A:
    attrs: ClassVar[list[str]]

class B:
    attrs: ClassVar[list[str]]

# TypeVar whose upper bound is a `type[...]` form.
T = TypeVar("T", bound=type[A] | type[B])

def f(cls: T) -> None:
    cls.attrs           # error[unresolved-attribute] — but `attrs` is on both arms

V = TypeVar("V", bound=type[A])

def g(cls: V) -> None:
    cls.attrs           # error[unresolved-attribute]

# --- Controls that resolve correctly ---

def plain(cls: type[A]) -> None:
    cls.attrs           # ok: plain type[A] parameter (no TypeVar)

W = TypeVar("W", bound=A | B)

def instance_bound(x: W) -> None:
    x.attrs             # ok: instance-union bound (fixed in 0.0.43, ruff #25561)
```

### Actual output (ty 0.0.49)

```
error[unresolved-attribute]: Object of type `T@f` has no attribute `attrs`
  --> repro.py:13:5
   |
13 |     cls.attrs
   |     ^^^^^^^^^

error[unresolved-attribute]: Object of type `V@g` has no attribute `attrs`
  --> repro.py:19:5
   |
19 |     cls.attrs
   |     ^^^^^^^^^

Found 2 diagnostics
```

### Expected

No diagnostics. `attrs` is declared on `A` (and `B`), and the TypeVar's upper
bound guarantees `cls` is `type[A]` (or `type[A] | type[B]`), so `cls.attrs`
should resolve exactly as it does for the `plain(cls: type[A])` control.
mypy and pyright both accept all four functions.

### Notes

- Likely related to ruff #25561 (treat instance-union-bound typevars like unions)
  and #24793 (attribute access on Callable-bounded TypeVars); this is the
  `type[...]`-bound counterpart.
- ty version: 0.0.49

### Environment

- ty 0.0.49

mvanhorn:
## Summary
Accessing a class attribute on a value whose static type is a TypeVar with a `type[...]` upper bound (e.g. `def f[T: type[C]](cls: T)` or a `type[A] | type[B]` union bound) wrongly emits `unresolved-attribute`, even though the attribute is declared on the bound class(es).

Fixes #3782


AlexWaygood:
Thanks. However, the source code for ty is found in the Ruff repo; we bundle Ruff as a submodule here and update that submodule every time we release ty. So your updates here would just be overwritten the next time we cut a release. Please feel free to submit a PR fixing this issue over at Ruff instead :-)

PR #3795, opened 2026-06-17
fix: Detect generator expressions containing await as async generators

### LINKED ISSUE #3723: `(x for x in xs if await ..)` not detected as an async generator
Generator expressions that ["contain `await` expressions"](https://docs.python.org/3/reference/expressions.html#generator-expressions) are async generators, but ty doesn't recognize them as such:

```py
import asyncio
from collections.abc import AsyncIterator


async def is_even(value: int) -> bool:
    return value % 2 == 0


async def main() -> None:
    generator = (value for value in [1, 2, 3] if await is_even(value))

    await anext(generator, None)  # ty: Argument to function `anext` is incorrect: Expected `SupportsAnext[Unknown]`, found `GeneratorType[int, None, None]`
```

mvanhorn:
## Summary
A generator expression that contains an `await` in a filter or yielded element (e.g. `(x for x in xs if await pred(x))`) is an asynchronous generator at runtime, but ty infers it as a synchronous `GeneratorType`, so `anext()` on it is wrongly rejected.

Fixes #3723


AlexWaygood:
Please see my comment at https://github.com/astral-sh/ty/pull/3794#issuecomment-4731751270

PR #3796, opened 2026-06-17
feat: Raise a diagnostic for duplicate ParamSpec component arguments at call sites

### LINKED ISSUE #3745: Raise diagnostic for duplicate `ParamSpec` components arguments
Ref https://github.com/astral-sh/ty/issues/1861

The following should raise a diagnostic given that `*args` and `**kwargs` are provided multiple times:

```py
from typing import Callable

def duplicate1[**P](f: Callable[P, None], *args: P.args, **kwargs: P.kwargs) -> None:
    f(*args, *args, **kwargs)          # should be rejected
    f(*args, **kwargs, **kwargs)       # should be rejected
```

Note that this is specific to when `ParamSpec`s are involved and should _not_ be raised for other non-ParamSpec cases like the following:

```py
def gradual(*args, **kwargs): ...

def works(*args, **kwargs) -> None:
    gradual(*args, *args, **kwargs)
    gradual(*args, **kwargs, **kwargs)
```

mvanhorn:
## Summary
When calling a callable typed with a `ParamSpec`, supplying `*args`/`**kwargs` more than once (e.g. `f(*args, *args, **kwargs)` or `f(*args, **kwargs, **kwargs)` where `args: P.args` and `kwargs: P.kwargs`) is invalid and should raise a diagnostic, because a `ParamSpec` represents an exact parameter list and the components may each be forwarded only once.

Fixes #3745


AlexWaygood:
Please see my comment at https://github.com/astral-sh/ty/pull/3794#issuecomment-4731751270

PR #3798, opened 2026-06-17
Improve the "Does ty have a strict mode?" FAQ

AlexWaygood:
## Summary

Stacked on top of https://github.com/astral-sh/ty/pull/3783

When users ask "Does ty have a strict mode?", they're not really asking "Does ty have a flag called `--strict`?". What they really mean is, "How do I get ty to be as strict as mypy/pyright in strict mode?". This PR reworks the FAQ entry to more clearly inform users the various options we'd recommend enabling if they want ty to apply strict type-checking to their codebase.

### DIFF
--- a/docs/coming-from-mypy-or-pyright.md
+++ b/docs/coming-from-mypy-or-pyright.md
@@ -21,7 +21,51 @@ This guide helps you migrate a project from
     corresponding to mypy's `check_untyped_defs` setting. The equivalent pyright setting is
     `analyzeUnannotatedFunctions = true`.
 
-## How to read this table
+## Stricter checking with ty
+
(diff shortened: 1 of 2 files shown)


carljm (approved):

sharkdp (approved):
This is great, thank you.

I think we should move the long form version of this to the migration page as a new section, and then link to that section from the FAQ entry.

The only other comment I have is that this is a rather long description (which is good). So maybe we could have a more clearly separated sub-section/admonition/table/bullet-point-list which summarizes for readers: "Here are the concrete steps that we suggest you perform to emulate mypy/pyright strict mode", even if that includes opinionated advise.

PR #3799, opened 2026-06-17
Update benchmarks for ty 0.0.50

charliermarsh:
## Summary

This PR refreshes our `BENCHMARKS.md` using ty 0.0.50, Pyrefly 1.1.0, Pyright 1.1.410, and mypy 2.1.0. All benchmarks were run on my M3 Max with 128 GB RAM and 16 cores.

### DIFF
--- a/BENCHMARKS.md
+++ b/BENCHMARKS.md
@@ -2,10 +2,10 @@
 
 All benchmarks were computed on macOS (Apple M4 Pro 14, 48 GB) with the following tool versions:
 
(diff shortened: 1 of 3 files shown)


BENCHMARKS.md
```diff
@@ -2,10 +2,10 @@
 
 All benchmarks were computed on macOS (Apple M4 Pro 14, 48 GB) with the following tool versions:
```
AlexWaygood:
```suggestion
All benchmarks were computed on macOS (Apple M3 Max 16, 128 GB) with the following tool versions:
```

Issue #3756, opened 2026-06-13
new rule - report an error on `Divergent` types


DetachHead:
as far as i can tell, `Divergent` is treated the same as `Any`, so ideally it should be an error (related: #622):

```py
from typing import reveal_type

def some_condition() -> bool:
    ...

x = [1]
while some_condition():
  x = [x]

value = x[0]

if not isinstance(value, int):
    value.asdf # no error
```

[Playground](https://play.ty.dev/49bd16c4-7b67-4fde-b9fa-d3adddb444f5)

there also doesn't seem to be any docs on what `Divergent` means. if it's the same as `Unknown`/`Any`, i would expect it to be documented [here](https://docs.astral.sh/ty/reference/typing-faq/#what-is-the-unknown-type-and-when-does-it-appear)

pyrefly's [`non-convergent-recursion`](https://pyrefly.org/en/docs/error-kinds/#non-convergent-recursion) rule catches this

AlexWaygood:
Arguably any time `Divergent` appears in a user-visible type, that's a ty limitation that we should aim to fix :-)

ibraheemdev:
> Arguably any time `Divergent` appears in a user-visible type, that's a ty limitation that we should aim to fix :-)

Hmm... I think `Divergent` can be useful signal to users for catching infinite recursion. What would you expect the type of `x` to be here?
```py
def flag() -> bool: ...

x = []
while flag():
  x.append(x)

reveal_type(x)  # revealed: list[Divergent]
```

carljm:
I agree with @ibraheemdev -- `Divergent` is often inherent to the types, not a ty inference limitation. And I'm not sure it's better for users if we just flatten it to `Unknown` vs explicitly signaling that there is non-convergent recursion happening. In other words, I don't consider appearance of `Divergent` in a user-visible type to represent a ty limitation we should fix.

I do think having a diagnostic in addition to the `Divergent` type, and documenting `Divergent`, would both make sense.

AlexWaygood:
Makes sense

carljm:
I've put up a PR for the documentation piece; I think the diagnostic is a nice-to-have for stable release, so putting this issue in 1.1 milestone.

sharkdp:
> I agree with [@ibraheemdev](https://github.com/ibraheemdev) -- `Divergent` is often inherent to the types, not a ty inference limitation.

[...]

PR #3809, opened 2026-06-18
Document ty's Divergent type in the typing FAQ

carljm:
## Summary

Add a new FAQ entry explaining what `Divergent` means in ty's type system and how it can appear, using a loop example.

Refs https://github.com/astral-sh/ty/issues/3756

## Testing

Docs build.

### DIFF
--- a/docs/reference/typing-faq.md
+++ b/docs/reference/typing-faq.md
@@ -46,6 +46,36 @@ msg.data = {"color": "blue"}
 
 ([Full example in the playground](https://play.ty.dev/862941a8-a3f6-4818-9ea1-d9d59b0bd2fa))
 
+## What is the `Divergent` type and when does it appear?
+
+`Divergent` is ty's way of representing type-level recursion that does not converge. Type inference
+can be recursive; for example, the type of a variable at the end of a loop can depend on its type at
+the beginning. ty analyzes such cycles repeatedly, looking for a stable result. If each iteration
+produces a new type, ty replaces the non-convergent part with `Divergent`.
+
+For example, each iteration of this loop wraps `x` in another list:
+
+```py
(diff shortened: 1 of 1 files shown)


docs/reference/typing-faq.md
```diff
@@ -46,6 +46,36 @@ msg.data = {"color": "blue"}
...
+reveal_type(x)  # Literal[1] | list[Divergent]
+```
+
+After the first analysis of the loop, `x` can be `Literal[1]` or `list[Literal[1]]`. The next
+analysis adds `list[list[Literal[1]]]`, and every subsequent analysis adds another level of nesting.
+This does not converge to a finite type. The revealed type preserves the known base case and uses
+`Divergent` for the infinitely expanding part.
+
+Like `Any` and `Unknown`, `Divergent` is a gradual type, so ty allows any operation on the
+`Divergent` part of a type. Unlike `Unknown`, it does not represent missing type information. It is
```
dhruvmanila, line 76: We could possibly provide a code snippet for this part explaining what it means to "allow any operation on the `Divergent` part of a type" like considering the above example:

```py
from typing import Literal

def first(a: Literal[1] | list[str]): ...
def second(a: Literal[1] | tuple[int]): ...

# works
first(x)

# Argument to function `foo` is incorrect: Expected `Literal[1] | tuple[int]`, found `Literal[1] | list[Divergent]` (invalid-argument-type)
second(x)
```

dhruvmanila (approved):

docs/reference/typing-faq.md
carljm: Hmm, I think this might be too much -- I think saying "Like `Any` and `Unknown`..." is probably enough?

Issue #3753, opened 2026-06-12
`call-non-callable` false positive after `hasattr()` narrowing


mxr:
### Summary

Hi, I'm seeing an issue with `hasattr` narrowing. `hasattr` narrowing infers the attribute type as `object` instead of `Unknown`, so attribute calls fail with `call-non-callable`. See repro at https://play.ty.dev/8c9a870a-67b5-4ea9-85a8-ca6d31158f84

```py
import unittest


def example(instance: unittest.TestCase) -> None:
    if hasattr(instance, "make_instance"):
        instance.make_instance(include_optional=False)  # Object of type `object` is not callable (call-non-callable) [Ln 6, Col 9]
```

`mypy` narrowing shows the attribute is `Any`, not `object` (which I believe is the correct behavior) and supports calling the attribute.

**Expected**: no error (attribute type should be `Unknown`)
**Actual**: `call-non-callable` — `object` is not callable

I believe the fix would be something like

```diff
# in types/narrow.rs
-                    // Since `hasattr` only checks if an attribute is readable,
-                    // the type of the protocol member should be a read-only property that returns `object`.
+                    // Since `hasattr` only checks if an attribute is readable and provides no
+                    // information about the attribute's type, use `Unknown` rather than `object`.
+                    // Using `object` prevents callers from calling the attribute.
                     let constraint =
-                        Type::protocol_with_readonly_members(self.db, [(attr, Type::object())]);
+                        Type::protocol_with_readonly_members(self.db, [(attr, Type::unknown())]);
```

### Version

ty 0.0.48 (10b243db0 2026-06-10)

sharkdp:
The `hasattr` narrowing behavior is intentionally strict. There is nothing that indicates that this attribute is callable. Even adding an additional `and callable(instance.make_instance)` check will not help, since that wouldn't give us any information about its signature.

You could either ignore this error (`# ty: ignore` or use a cast to a `Callable` type), or, if possible, restructure your code so that the `make_instance` attribute is typed, e.g. by subclassing `TestCase`. If you have multiple usage sites, you could consider doing something like the following, which wouldn't give you any more type safety, but move the problem to away from the usage site. And you could model the return type of `make_instance`.


```py
from typing import TYPE_CHECKING
import unittest

if TYPE_CHECKING:
    from typing import TypeIs, Protocol

    from ty_extensions import Intersection

    class HasMakeInstance(Protocol):
[...]

PR #3818, opened 2026-06-21
Behavioral differences docs

mxr:
## Summary

Doc updates in response to https://github.com/astral-sh/ty/issues/3753. I am by no means a typing expert and expect this section, if accepted, will evolve significantly over time.

## Test Plan

Ran `pre-commit` locally


sharkdp:
Thank you for your contribution.

As I said in the issue, this is currently not a priority, and I would only consider spending time on it (reviewing, adapting) if someone invests into writing a more general section about our behavioral differences to mypy/pyright in narrowing, not just for `hasattr` (which is much less commonly used compared to other forms of narrowing). Since all of this is currently in flux, I don't think it's a good idea to do this right now - sorry!

PR #3821, opened 2026-06-22
Document Python version support

### LINKED ISSUE #878: Do we support 3.7 or not?
In the [docs](https://docs.astral.sh/ty/reference/cli/#ty-check--python-version), it is said that you can supply `--python-version 3.7` but you cant pip install with python 3.7.

What is the intended outcome, can we support 3.7?

charliermarsh:
## Summary

Document that we officially support type checking code targeting Python 3.10 and later, while Python 3.7 through 3.9 remain available (but only with best-effort behavior).

Closes https://github.com/astral-sh/ty/issues/878.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,10 @@ ty uses `0.0.x` versioning. ty does not yet have a stable API; breaking changes,
 to diagnostics, may occur between any two versions. See the [type system support](https://github.com/astral-sh/ty/issues/1889)
 tracking issue for a detailed overview of currently supported features.
 
+ty officially supports type checking code that targets Python 3.10 and later. Earlier versions
+(Python 3.7 through 3.9) can still be selected, but may result in false negatives or false
+positives due to a lack of bundled standard library stubs.
+
 ## FAQ
 
 <!-- We intentionally use smaller headings for the FAQ items -->
--- a/docs/python-version.md
+++ b/docs/python-version.md
@@ -40,3 +40,10 @@ ty will fall back to the latest stable Python version supported by ty (currently
 The Python version may also be explicitly specified using the
 [`python-version`](./reference/configuration.md#python-version) setting or the
 [`--python-version`](./reference/cli.md#ty-check--python-version) flag.
+
+ty officially supports type checking code that targets Python 3.10 and later, though earlier
+versions (Python 3.7 through 3.9) can still be selected as targets. When analyzing code under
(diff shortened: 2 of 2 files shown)


docs/python-version.md
```diff
@@ -40,3 +40,10 @@ ty will fall back to the latest stable Python version supported by ty (currently
 The Python version may also be explicitly specified using the
 [`python-version`](./reference/configuration.md#python-version) setting or the
 [`--python-version`](./reference/cli.md#ty-check--python-version) flag.
+
+ty officially supports type checking code that targets Python 3.10 and later, though earlier
+versions (Python 3.7 through 3.9) can still be selected as targets. When analyzing code under
+`--python-version=3.7` or similar, ty will continue to check version-dependent syntax, perform
+narrowing based on `sys.version_info`, and more, but may produce false positives or false
+negatives for standard-library APIs as its bundled stubs do not fully describe those Python
+versions.
```
MichaReiser, line 49: Should we add a similar blurb to the `--python` and `environment.python` CLI and configuration options?

MichaReiser (approved):
Thank you

docs/python-version.md
charliermarsh: Yeah. I think this needs to go in Ruff, right?

MichaReiser: Yes

charliermarsh: Done in https://github.com/astral-sh/ruff/pull/26217.

README.md
```diff
@@ -75,6 +75,10 @@ ty uses `0.0.x` versioning. ty does not yet have a stable API; breaking changes,
 to diagnostics, may occur between any two versions. See the [type system support](https://github.com/astral-sh/ty/issues/1889)
 tracking issue for a detailed overview of currently supported features.
 
+ty officially supports type checking code that targets Python 3.10 and later. Earlier versions
+(Python 3.7 through 3.9) can still be selected, but may result in false negatives or false
+positives due to a lack of bundled standard library stubs.
```
AlexWaygood, line 80: this doesn't really feel like it belongs in the "Version policy" section? It doesn't have anything to do with ty's versioning

charliermarsh: Thought about this and decided it was fine / preferable to adding another top-level section.

AlexWaygood: it could work well in the FAQs section?

charliermarsh: I’ll try it out!

PR #3824, opened 2026-06-22
Document recent changes to how exit codes are determined

AlexWaygood:
## Summary

Following https://github.com/astral-sh/ruff/pull/26157

### DIFF
--- a/docs/reference/exit-codes.md
+++ b/docs/reference/exit-codes.md
@@ -13,7 +13,7 @@ ty supports three command line arguments that change how exit codes work:
 
 - `--exit-zero`: ty will exit with `0` even if violations were found.
 - `--error-on-warning`: ty will exit with `1` if it finds any violations with severity `warning` or
-    higher. This is usually the default behaviour, but can be overridden using a configuration file.
+    higher.
 - `--exit-zero-on-warning`: ty will only exit with `1` if it finds violations with severity `error` or higher.
 
 `--error-on-warning` cannot be supplied at the same time as `--exit-zero` or `--exit-zero-on-warning`.
--- a/docs/reference/exit-codes.md
+++ b/docs/reference/exit-codes.md
@@ -2,15 +2,18 @@
 
 The ty command line interface uses the following exit codes:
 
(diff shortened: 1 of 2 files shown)


docs/reference/exit-codes.md
```diff
@@ -2,15 +2,18 @@
...
+| `2`       | invalid CLI options, invalid configuration, or IO errors   |
+| `101`     | internal error                                             |
 
-ty supports two command line arguments that change how exit codes work:
+ty supports three command line arguments that change how exit codes work:
 
 - `--exit-zero`: ty will exit with `0` even if violations were found.
 - `--error-on-warning`: ty will exit with `1` if it finds any violations with severity `warning` or
-    higher.
+    higher. This is usually the default behaviour, but can be overridden using a configuration file.
```
sharkdp: This sounds a bit like the configuration file would overwrite the CLI argument.

docs/reference/exit-codes.md
```diff
@@ -2,15 +2,18 @@
...
+- `--exit-zero-on-warning`: ty will only exit with `1` if it finds at least one violation with severity `error` or higher.
```
sharkdp:
```suggestion
- `--exit-zero-on-warning`: ty will only exit with `1` if it finds violations with severity `error` or higher.
```

sharkdp (approved):
Thank you

docs/reference/exit-codes.md
AlexWaygood: hmm, yeah. I was trying to provide a justification for why the CLI flag might sometimes be useful (if you want to temporarily override configuration provided by a config file)... but maybe this is just confusing and I should get rid of the second sentence?

sharkdp: I think it's fine to say "This is the default behavior." without justifying why that option exists nevertheless, yes.

AlexWaygood: Micha objected to me saying "This is the default behaviour" in my Ruff PR without qualifying that the default can be overridden by configuration files, so I'll just get rid of this second sentence entirely

PR #4038, opened 2026-07-19
Update documentation on mypy/pyright `--strict` parity, and achieving maximum strictness with ty

AlexWaygood:
## Summary

Update the documentation to mention:
- The new `blanket-ignore-comment` rule (and its mypy/basedpyright equivalents)
- The new `strict-literal-narrowing` analysis setting
- Several other "noisier" rules that folks who truly want maximum strictness might theoretically want to enable

With these changes, the only disabled-by-default rule that is still unmentioned in the "making ty stricter" section of these docs is https://docs.astral.sh/ty/reference/rules/#missing-override-decorator. I chose not to mention it here because it doesn't really feel like a _strictness_ issue to me. While I can understand why some codebases might want to enforce that `@override` should be applied to all method overrides, it has no impact on soundness if the decorator is omitted.

## Test Plan

`uvx prek run -a`

### DIFF
--- a/docs/coming-from-mypy-or-pyright.md
+++ b/docs/coming-from-mypy-or-pyright.md
@@ -54,10 +54,33 @@ preview = true
 
 This configuration:
 
-- Ensures ty exits with a non-zero status if it emits any warning-level diagnostics
 - Enables ty's disabled-by-default [`missing-type-argument`](https://docs.astral.sh/ty/reference/rules/#missing-type-argument) and [`possibly-unresolved-reference`](https://docs.astral.sh/ty/reference/rules/#possibly-unresolved-reference) rules
 - Extends Ruff's default rules with the [`ANN`](https://docs.astral.sh/ruff/rules/#flake8-annotations-ann) and [`PYI`](https://docs.astral.sh/ruff/rules/#flake8-pyi-pyi) rule categories, both of which are focussed on type-annotating your code more effectively
 - Enables Ruff's preview mode so that `PYI033` also checks `.py` files
+
+An even stricter configuration -- that goes beyond what mypy and pyright check for in their default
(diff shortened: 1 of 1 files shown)


docs/coming-from-mypy-or-pyright.md
```diff
@@ -54,11 +54,34 @@ preview = true
 
 This configuration:
 
-- Ensures ty exits with a non-zero status if it emits any warning-level diagnostics
 - Enables ty's disabled-by-default [`missing-type-argument`](https://docs.astral.sh/ty/reference/rules/#missing-type-argument) and [`possibly-unresolved-reference`](https://docs.astral.sh/ty/reference/rules/#possibly-unresolved-reference) rules
 - Extends Ruff's default rules with the [`ANN`](https://docs.astral.sh/ruff/rules/#flake8-annotations-ann) and [`PYI`](https://docs.astral.sh/ruff/rules/#flake8-pyi-pyi) rule categories, both of which are focussed on type-annotating your code more effectively
 - Enables Ruff's preview mode so that `PYI033` also checks `.py` files
 
+An even stricter configuration -- that goes beyond what mypy and pyright check for in their default
```
charliermarsh, line 61: (I think it should be stylized as "Pyright" everywhere.)

charliermarsh (approved):

docs/coming-from-mypy-or-pyright.md
charliermarsh: (I think I'd also expect mypy to be title-cased when at the start of a sentence like the mypy site, but defer to you on that one.)

AlexWaygood: Mypy should be title-cased at the start of sentences but lowercased elsewhere, yep

PR #4154, opened 2026-08-01
ci: add names to cargo-dist release steps for readability

MGPOCKY:
## Problem

In `.github/workflows/release.yml`, long cargo-dist `run` steps had an `id` but no `name:`, so the Actions UI shows generic labels for plan / global build / host.

## Changes

Semantics are unchanged: same step `id`s, same `GITHUB_OUTPUT` keys, and the same job/step output wiring.

- Add `name: Plan release with cargo-dist` to `id: plan`
- Add `name: Build global cargo-dist artifacts` to `id: cargo-dist`
- Add `name: Host release with cargo-dist` to `id: host`

## Test plan
- [ ] Confirm `plan.outputs.val` still comes from `steps.plan.outputs.manifest`
- [ ] Confirm global upload still uses `steps.cargo-dist.outputs.paths`
- [ ] Confirm `host.outputs.val` still comes from `steps.host.outputs.manifest`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -101,7 +101,8 @@ jobs:
       # functionality based on whether this is a pull_request, and whether it's from a fork.
       # (PRs run on the *source* but secrets are usually on the *target* -- that's *good*
       # but also really annoying to build CI around when it needs secrets to work right.)
-      - id: plan
+      - name: Plan release with cargo-dist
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Thanks, but this file is generated, and even if it weren't, I don't think this counts as a meaningful improvement unfortunately

Issue #4157, opened 2026-08-02
Overriding `__hash__` does not make the object unhashable; `list` is also `Hashable`?


cmp0xff:
### Summary

https://play.ty.dev/1a18dc5a-253d-4f33-a6ee-7faee6a4a5c1

```py
from collections.abc import Hashable
from typing import ClassVar


class DataFrame:
    __hash__: ClassVar[None]

def f(_: Hashable) -> None:
    pass

# ty is okay; mypy, pyright and pyrefly raise
f(DataFrame())
f(["1"])
```

### Version

playground

carljm:
Currently we consider the type `Hashable` equivalent to the type `object`, which is technically correct, since `object` specifies a non-None `__hash__`, and thus it should be a Liskov violation to ever create a non-hashable Python object. But obviously this doesn't accommodate very well the reality of the (unsound) way that `Hashable` and hashability are used in real-world Python type annotations.

I think we should not consider `Hashable` equivalent to `object`, and we should actually validate the `__hash__` type of the provided type against `Hashable`.

Issue #3781, opened 2026-06-16
Flag non-hashable objects as error in `set`/`dict`


Werni2A:
### Summary

`ty` does not flag when non-hashable objects are inserted into `set` or `dict` which causes runtime errors. Below is an example with `set`.

Working example with a hashable object.

[Playground Link](https://play.ty.dev/cfccd0e2-986c-47a2-bfff-15bc3d74cef3)

```python
from dataclasses import dataclass

@dataclass(frozen=True)
class HashableObject:
    content: str

obj = HashableObject("content")
print({obj})
```

Failing example with a non-hashable object where it fails during runtime:

[Playground Link](https://play.ty.dev/f38c32dd-5de4-4e90-bca2-5d49f0e73408)

```python
from dataclasses import dataclass

@dataclass
class NonHashableObject:
    content: str

obj = NonHashableObject("content")
print({obj})
```

runtime error:

```text
Failed to run Python script: PythonError: Traceback (most recent call last):
  File "/lib/python313.zip/_pyodide/_base.py", line 523, in eval_code
    .run(globals, locals)
     ~~~^^^^^^^^^^^^^^^^^
  File "/lib/python313.zip/_pyodide/_base.py", line 357, in run
    coroutine = eval(self.code, globals, locals)
  File "main.py", line 8, in <module>
    print({obj})
          ^^^^^
TypeError: unhashable type: 'NonHashableObject'
```

Might be related to https://github.com/astral-sh/ty/issues/1162

### Version

4f7fc885d

sharkdp:
Thank you for reporting this.

The [`set` constructor in typeshed](https://github.com/python/typeshed/blob/5e927a88f7c37b2592371a419cf4dd355a3ae4b7/stdlib/builtins.pyi#L1427-L1430) does not indicate that elements must be `Hashable`, so I guess that would have to be an explicit rule in ty?

AlexWaygood:
In general we don't try to model hashability in typeshed, because other type checkers also have difficulties with the `Hashable` protocol and the way so many classes violate Liskov w.r.t. their `__hash__` attributes (though other type checkers generally do not have this problem as "severely" as ty).

PR #4165, opened 2026-08-03
Update mypy/pyright comparison table

jelle-openai:
Expanded the table with additional error codes from all three checkers.

This is going to be hard to review, sorry.


AlexWaygood:
I hate markdown tables, I hate markdown tables, I hate markdown tables

AlexWaygood:
Two things spotted by codex:

- `reportTypedDictNotRequiredAccess` is newly mapped to `invalid-assignment`, but another row simultaneously says ty has no equivalent for that same diagnostic. Those claims contradict each other.
- The new `reportUnhashable` row says the feature is tracked by [#4157](https://github.com/astral-sh/ty/issues/4157), but that issue is already closed as completed.

AlexWaygood:
A more full review from codex:

1. `invalid-assignment` → `reportTypedDictNotRequiredAccess` is incorrect. That Pyright diagnostic concerns reading a non-required TypedDict key, not assignment. ty currently emits nothing for that case, which the table already acknowledges in another row. Remove this mapping from `invalid-assignment`.

2. `ANN401` is not limited to function parameters. It also flags return annotations:

   ```python
   from typing import Any

   def f() -> Any: ...
   ```

   Change “function parameters only” to “function annotations only”.

3. The `ANN401`/`explicit-any` row omits basedpyright’s direct equivalent: `reportExplicitAny`. Add it with a “basedpyright only” qualifier.

4. The `reportUnhashable` row links to the wrong tracking issue. #4157 is already closed and concerns `Hashable` protocol behavior. The actual open issue for unhashable set elements and dictionary keys is #3781.

jelle-openai:
@AlexWaygood could you take another look?

docs/coming-from-mypy-or-pyright.md
```diff
@@ -101,98 +104,158 @@ table below for more details.
...
+| None yet for instantiating abstract classes                                                                                  | [`abstract`][mypy-abstract]                                                                                                    | [`reportAbstractUsage`][reportabstractusage]                                                                                                                                           |
+| None yet for passing abstract classes where concrete classes are required                                                    | [`type-abstract`][mypy-type-abstract]                                                                                          |                                                                                                                                                                                        |
```
AlexWaygood: I think we intend at some point to implement a broader rule that complains for _any_ instantiation of a `type[]` type, but I doubt we will ever implement a 1:1 mapped rule for mypy's `type-abstract` rule. So it's more of a "none planned" than "none yet" situation

docs/coming-from-mypy-or-pyright.md
```diff
@@ -101,98 +104,158 @@ table below for more details.
...
+| [`unresolved-reference`][ty-unresolved-reference], [Ruff `F823`][ruff-f823]                                                  | [`name-defined`][mypy-name-defined]<br>[`used-before-def`][mypy-used-before-def]                                               | [`reportUndefinedVariable`][reportundefinedvariable]<br>[`reportUnboundVariable`][reportunboundvariable]                                                                               |
+| [`unsound-assignment`][ty-unsound-assignment] (variables only)                                                               |                                                                                                                                | [`reportAny`][reportany] (basedpyright only; broader check)                                                                                                                            |
```
AlexWaygood: I'm not sure I'd really to `reportAny` -- that rule is far more sweeping, and I'm not sure we ever intend to implement something as aggressive as that rule

AlexWaygood (approved):
LG, two optional comments

PR #4202, opened 2026-08-06
Bump version to 0.0.69

sharkdp:


CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
...
+
+    This new behavior is more similar to what other major type checkers do. For example:
+
+    ```py
+    def f(xs: object):
+        if isinstance(xs, list):
+            reveal_type(xs)  # now: `list[Unknown]`, before: `Top[list[Unknown]]`
+    ```
+
+    **Note**: you might consider enabling [`analysis.strict-generic-narrowing`](https://docs.astral.sh/ty/reference/configuration/#strict-generic-narrowing)
```
sharkdp, line 19: this link doesn't exist yet, but hopefully will, soon after the release

dhruvmanila (approved):
Looks good, ~~just want to confirm that the relative links works as expected in the rendered docs?~~

Edit: tested the relative links works as expected

sharkdp:
> Looks good, ~just want to confirm that the relative links works as expected in the rendered docs?~

ah sorry, I tested those! And I think maybe the docs build would even error if they wouldn't work?

PR #4213, opened 2026-08-07
Enable PGO for Linux x86-64 ty releases

charliermarsh:
## Summary

This PR enables PGO for ty releases, starting with Linux x86-64. The approach follows that outlined in https://github.com/astral-sh/ruff/pull/27570.

### Design

The release pipeline is modified as follows:

- We build an instrumented release binary using the Rust toolchain pinned by the Ruff submodule.
- We run `ty check` on the same ten pinned ecosystem projects used by Ruff. In total, the corpus contains 618 Python and stub files.
- We also exercise `ty server` against a temporary project, covering diagnostics, hover, go-to-definition, completion, and incremental cross-file edits.
- We merge the profiles via `llvm-profdata`.
- We derive LLVM's hot-code threshold from the profile's 95th percentile, matching its size-optimization threshold and avoiding unnecessary expansion of moderately hot functions. (Without this, performance was marginally better, but wheel size _increased_ by 5-10%!)
- We feed the result back into the existing `maturin` build.

### Results

Here's the initial, untuned PGO performance on seven held-out projects (and language-server performance on three held-out projects):

| Held-out project | `ty check` | Incremental edits |
| --- | ---: | ---: |
| Black | 7.1% faster | 17.1% faster |
| isort | 13.3% faster | 18.1% faster |
| Jinja | 5.5% faster | 13.7% faster |
| Django | 25.5% faster | — |
| pandas | 15.9% faster | — |
| scikit-learn | 19.6% faster | — |
| SymPy | 21.5% faster | — |
| Geometric mean | **15.8% faster** | **16.3% faster** |

As an aside, profiling the language server explicitly (i.e., including the language server commands in the PGO training) improved incremental-edit latency by another 7.6% over CLI-only PGO; without that training, incremental edit performance was _still_ up (but not quite as much).

Beyond raw performance, with the tuned PGO settings, we see the following results:

- **Binary size decreased by 7.98%** (28.10 MB to 25.86 MB compared with non-PGO).
- **Wheel size decreased by 1.32%** (12.71 MB to 12.54 MB compared with non-PGO), and by 14.11% compared with untuned PGO.
- **Release archive size decreased by 1.53%** (12.35 MB to 12.16 MB compared with non-PGO), and by 14.25% compared with untuned PGO.
- **Incremental language-server edits are 3.30% faster than untuned PGO**.
[...]

PR #4216, opened 2026-08-07
Enable PGO for macOS ARM64 ty releases

charliermarsh:
## Summary

This PR enables PGO for ty's macOS ARM64 releases, following the approach outlined in https://github.com/astral-sh/ty/pull/4213. The existing linker configuration and safe identical-code folding are preserved.

Across seven held-out macOS projects, `ty check` finishes 5.0% faster and uses 7.1% less CPU; incremental language-server edits are 7.3% faster. The production executable is 3.6% smaller than without PGO, although the wheel and release archive are each 2.8% larger.

The macOS ARM64 release job uses the same Namespace runner as Ruff, reducing its full-stack PGO build from 23m 47s to 6m 39s (72%). Three Namespace builds finished in 6m 38s to 7m 59s; the previous Depot runner took 11m 20s.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -6,5 +6,6 @@ self-hosted-runner:
   labels:
     - depot-ubuntu-24.04-4
     - depot-ubuntu-24.04-8
+    - namespace-profile-macos-15
     - github-windows-2025-x86_64-8
     - github-windows-2025-x86_64-16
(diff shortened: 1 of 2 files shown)


MichaReiser:
Can you please add benchmark results that go beyond wheel size?

charliermarsh:
Added.

MichaReiser (approved):

PR #4217, opened 2026-08-07
Enable PGO for Windows x86-64 ty releases

charliermarsh:
## Summary

This PR enables PGO for ty's Windows x86-64 releases, following the approach outlined in https://github.com/astral-sh/ty/pull/4213. Both builds preserve static Visual C++ runtime linking.

On the Windows holdout set, `ty check` uses 19.7% less CPU and finishes 17.5% faster, while incremental language-server edits get 14.7% faster. The release binary gets 9.6% smaller, and the wheel and release archive each get 3.4% smaller.

Only the Windows x86-64 PGO job uses the same 16-core Namespace runner as Ruff, reducing the full-stack build from 26m 08s to 12m 24s (53%). The previous 32-core Depot runner took 14m 17s. With Linux x86-64 at 11m 28s and Linux ARM64 at 11m 37s, Windows is now the release bottleneck. The remaining Windows targets keep their existing runners.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -7,6 +7,6 @@ self-hosted-runner:
     - depot-ubuntu-24.04-4
     - depot-ubuntu-24.04-8
     - namespace-profile-macos-15
-    - depot-windows-2025-32
+    - namespace-profile-windows-2022-x86-64-16x32
     - github-windows-2025-x86_64-8
     - github-windows-2025-x86_64-16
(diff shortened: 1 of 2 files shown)


MichaReiser (approved):
It would be great to add a more detailed table of the performance improvement, much like the one you added to the first PR, if you have it.

MichaReiser (approved):
I hope we don't run into any issues with dependencies not being available on windows.

PR #4218, opened 2026-08-07
Enable PGO for Linux ARM64 ty releases

charliermarsh:
## Summary

This PR enables PGO for ty's Linux ARM64 releases, following the approach outlined in https://github.com/astral-sh/ty/pull/4213. (To enable PGO, we also move to a native ARM64 runner.) The existing jemalloc page-size configuration is preserved.

Across four held-out Linux ARM64 projects, `ty check` uses 15.0% less CPU and finishes 11.9% faster. The release binary gets 4.9% smaller, the wheel gets 1.2% smaller, and the release archive gets 1.4% smaller.

The Linux ARM64 PGO build uses an eight-core Depot runner, reducing its runtime from 14m 51s on a four-core GitHub runner to 11m 37s (22%). In the matched full-stack run, eight-core Linux x86-64 finished in 11m 28s and Namespace Windows finished in 12m 24s, making Windows the release bottleneck.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -6,6 +6,7 @@ self-hosted-runner:
   labels:
     - depot-ubuntu-24.04-4
     - depot-ubuntu-24.04-8
+    - depot-ubuntu-24.04-arm-8
     - namespace-profile-macos-15
     - namespace-profile-windows-2022-x86-64-16x32
     - github-windows-2025-x86_64-8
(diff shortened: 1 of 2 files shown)


MichaReiser (commented):
Do I understand your summary correctly, that this does not improve performance?

charliermarsh:
No, that's not implied. I just didn't run the results since I need an ARM machine (much like the Windows build) and I'll have to do a dedicated pass to generate them.

charliermarsh:
Added (via GitHub Actions so YMMV!).

MichaReiser (approved):

PR #4224, opened 2026-08-10
Use Depot runners for release builds

charliermarsh:
Move source distributions and Linux release builds onto four-core Ubuntu 24.04 Depot runners, matching the CPU count of the GitHub-hosted runners they replace while avoiding the runner queues observed in recent releases. Linux cross builds use the same runner size because they have not generally been the release bottleneck. Register the self-hosted runner label with actionlint.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -6,6 +6,5 @@ self-hosted-runner:
   labels:
     - depot-ubuntu-24.04-4
     - depot-ubuntu-24.04-8
-    - depot-ubuntu-latest-4
     - github-windows-2025-x86_64-8
     - github-windows-2025-x86_64-16
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -4,6 +4,8 @@
 self-hosted-runner:
   # Various runners we use that aren't recognized out-of-the-box by actionlint:
   labels:
+    - depot-ubuntu-24.04-4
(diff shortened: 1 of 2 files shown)


.github/workflows/build-binaries.yml
```diff
@@ -227,7 +227,7 @@ jobs:
 
   linux:
     if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-build') }}
-    runs-on: ubuntu-latest
+    runs-on: depot-ubuntu-latest-4
```
ewdurbin: looks like we pinned most of the other actions to 24.04, any reason in particular not to pin this one while we're at it?

charliermarsh: I think because it was on `ubuntu-latest` already, but should change it

zanieb: +1 on pinning

charliermarsh:
Done

.github/actionlint.yaml
```diff
@@ -4,6 +4,8 @@
 self-hosted-runner:
   # Various runners we use that aren't recognized out-of-the-box by actionlint:
   labels:
+    - depot-ubuntu-24.04-4
+    - depot-ubuntu-24.04-8
     - depot-ubuntu-latest-8
```
charliermarsh: I'm going to remove the rest of these in a subsequent PR

PR #4235, opened 2026-08-11
Use an eight-core Ubuntu 24.04 runner for prek CI

charliermarsh:
## Summary

Previously, we ran `prek` on a 16-core Ubuntu 22.04 Depot runner. We now use an 8-core Ubuntu 24.04 runner and remove the unused Depot runner labels from the actionlint configuration.

| Runner | `prek` job | `Run prek` step | Approximate core-seconds |
| --- | ---: | ---: | ---: |
| 16 cores | 46s | 10s | 736 |
| 8 cores | 68s | 28s | 544 |
| 4 cores | 95s | 53s | 380 |

The 8-core runner uses approximately 26% fewer core-seconds than the previous 16-core runner while finishing 27 seconds sooner than the 4-core alternative. `prek` remains outside the overall CI critical path, where `python package` finished last in all eight recent completed runs on `main`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,7 +49,7 @@ jobs:
 
   prek:
     name: "prek"
-    runs-on: depot-ubuntu-24.04-4
+    runs-on: depot-ubuntu-24.04-8
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,7 +49,7 @@ jobs:
 
   prek:
     name: "prek"
-    runs-on: depot-ubuntu-24.04-16
+    runs-on: depot-ubuntu-24.04-8
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,7 +49,7 @@ jobs:
 
   prek:
     name: "prek"
-    runs-on: depot-ubuntu-22.04-16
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
 
   prek:
     name: "prek"
-    runs-on: depot-ubuntu-22.04-16
+    runs-on: depot-ubuntu-24.04-16
```
MichaReiser: Lol, 16 cores to run prek? Well, this is pre-existing but feels a bit extreme. what do we run in prek

MichaReiser (approved):

.github/workflows/ci.yaml
charliermarsh: Lol I had the same thought. I'm benching it now. Seems silly.

charliermarsh:
I changed it to `-4`.

MichaReiser (approved):

AlexWaygood:
that seems to increase the runtime from 10s to 53s? That's a lot longer to find out that you need to reformat some markdown files to get CI green

AlexWaygood:
28s seems fine, I'm happy to compromise on that 😄

MichaReiser:
It should be 0s seconds if you use the pre-commit hook ;)

charliermarsh:
Alright we can go to `-8`, I just haven't had this fail in a long time anyway since I have Codex do the release PRs haha.

MichaReiser:
I don't know which prek checks are expensive, and i don't know prek well enough, but, if it's important, we could run the "cheap" checks first.

AlexWaygood:
Oh, this is the ty repository... oops. I guess I feel about it much less strongly here than over at Ruff haha. But changing anything will also buy us less here, since PRs to the ty repository are much less frequent than PRs to the Ruff repository

AlexWaygood:
> I don't know which prek checks are expensive, and i don't know prek well enough, but, if it's important, we could run the "cheap" checks first.

they run concurrently in batches, and I think we already more-or-less do the cheap batches first

PR #4237, opened 2026-08-11
Enable PGO for Linux x86-64 ty releases via Ruff

charliermarsh:
## Summary

This is an alternative to https://github.com/astral-sh/ty/pull/4213 that keeps the PGO training script in Ruff, where it shares ecosystem-project setup with mypy-primer and runs continuously against ty changes. It depends on https://github.com/astral-sh/ruff/pull/27654.

The ty release workflow calls `ruff/scripts/build_ty_pgo.py --train-only` for Linux x86-64, then passes the generated profile and profile-derived hotness threshold into the existing `maturin` build. The shared script preserves the ten-project ecosystem corpus, installed project dependencies, project-specific Python versions, and incremental language-server training.

The complete ty-side change is the release workflow and Ruff submodule update; no PGO Python script lives in this repository.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -16,6 +16,7 @@ on:
     paths:
       # When we change pyproject.toml, we want to ensure that the maturin builds still work.
       - pyproject.toml
+      - ruff
       # And when we change this workflow itself...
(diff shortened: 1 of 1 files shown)


charliermarsh:
Gonna go with the in-ty pipeline for now.

PR #4266, opened 2026-08-14
Align Renovate config with preset

woodruffw:
## Summary

See https://github.com/astral-sh/uv/pull/21142 for context. 

## Test Plan

NFC.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,37 +1,9 @@
 {
   $schema: "https://docs.renovatebot.com/renovate-schema.json",
-  dependencyDashboard: true,
-  suppressNotifications: ["prEditedNotification"],
   extends: ["github>astral-sh/renovate-config"],
-  labels: ["internal"],
   schedule: ["before 4am on Wednesday"],
-  semanticCommits: "disabled",
-  separateMajorMinor: false,
-  prHourlyLimit: 10,
   enabledManagers: ["github-actions", "pre-commit", "pep621", "custom.regex"],
-  "pre-commit": {
-    enabled: true,
-  },
   packageRules: [
-    // Pin GitHub Actions to immutable SHAs.
-    {
-      matchDepTypes: ["action"],
-      pinDigests: true,
-    },
-    // Annotate GitHub Actions SHAs with a SemVer version.
-    {
-      extends: ["helpers:pinGitHubActionDigests"],
-      extractVersion: "^(?<version>v?\\d+\\.\\d+\\.\\d+)$",
-      versioning: "regex:^v?(?<major>\\d+)(\\.(?<minor>\\d+)\\.(?<patch>\\d+))?$",
-    },
-    {
-      // Group upload/download artifact updates, the versions are dependent
-      groupName: "Artifact GitHub Actions dependencies",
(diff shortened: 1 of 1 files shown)


AlexWaygood:
codex reckons this changes behaviour in a few small ways. (1) looks like it might be fine (not sure I fully understand it, haven't dug in deeply) but (2-3) look possibly unintentional?

---

1. **The set of dependencies Renovate can update changes.** The deleted `packageRules` entry containing `extractVersion` and `versioning` has no matcher of its own, so those GitHub Actions-specific settings currently apply to every dependency. The shared preset correctly restricts them to actual GitHub Actions. Removing the local rule therefore changes version handling for Python runtimes, Python packages, pre-commit hooks, and maturin. Renovate fixed this exact upstream problem in https://github.com/renovatebot/renovate/pull/44830.

   | Dependency | Current version | Valid before | Valid after |
   |---|---|---|---|
   | Python runtime in the release workflow | `3.13` | No | Yes |
   | `validate-pyproject` pre-commit hook | `v0.25` | No | Yes |
   | `shellcheck-py` pre-commit hook | `v0.11.0.1` | No | Yes |

   Python packages also switch from the GitHub Actions-oriented regex to normal `pep440` versioning. The Renovate comment on this PR already predicts a newly enabled **“Update dependency python to 3.14”** PR. Removing the leaking rule probably fixes a bug, but it is not a no-op.

2. **The hourly PR limit changes from 10 to unlimited.** The repository currently sets `prHourlyLimit: 10`, while the shared preset sets [`prHourlyLimit: 0`](https://github.com/astral-sh/renovate-config/blob/79c32d43fb34a89d281502081efd58b60e03c65e/preset.json5#L63). Renovate interprets `0` as unlimited: https://docs.renovatebot.com/configuration-options/#prhourlylimit. If the existing limit is intentional, `prHourlyLimit: 10` needs to stay.

3. **Security PR titles and commit messages gain `[SECURITY]`.** The repository currently sets `vulnerabilityAlerts.commitMessageSuffix: ""`, but the shared preset does not override that field. Removing the repository setting restores Renovate's default `commitMessageSuffix: "[SECURITY]"`: https://docs.renovatebot.com/configuration-options/#vulnerabilityalerts. The security labels themselves remain unchanged.

woodruffw:
(3) was intentional, that's me trying to unify our handling across repos. But (2) was mostly unintentional, if you're worried about a surge in Renovate PRs then we can set that in the preset (IMO it probably should be anyways?)

AlexWaygood:
I think unlimited is probably what we want for our repos tbh, so (2) is probably a good change!

I remember getting rid of the scary "[SECURITY]" suffix was something I did in response to feedback from @zanieb shortly after we switched to renovate, though -- see https://discord.com/channels/1039017663004942429/1080996377611489350/1221875504026484847, and https://github.com/astral-sh/ruff/pull/10579

Anyway, I agree that all this is better handled in the org-wide preset

AlexWaygood (approved):
thanks!

woodruffw:
Thanks! And yeah thinking more I agree about the `[SECURITY]` prefix, I'll remove.

PR #4268, opened 2026-08-14
Bump version to 0.0.72

charliermarsh:


CHANGELOG.md
```diff
@@ -1,5 +1,48 @@
 # Changelog
 
+## 0.0.72
+
+Released on 2026-08-14.
+
+### Preview features
+
+- Check PEP 723 scripts in isolation ([#27462](https://github.com/astral-sh/ruff/pull/27462))
```
AlexWaygood, line 9: heh, I thought all of ty was in preview 😆

CHANGELOG.md
```diff
@@ -1,5 +1,48 @@
...
+### Library support
+
+- Pydantic: Accept documented boolean and fractional inputs ([#27754](https://github.com/astral-sh/ruff/pull/27754))
+- Pydantic: Accept enum members for lax string and integer fields ([#27751](https://github.com/astral-sh/ruff/pull/27751))
+
+### Diagnostics
+
+- Diagnose invalid module-level `__getattr__` calls ([#27507](https://github.com/astral-sh/ruff/pull/27507))
+- Improve assignability hints for protocols and `TypedDict` types ([#27717](https://github.com/astral-sh/ruff/pull/27717))
+- Report deprecated unary operations ([#27584](https://github.com/astral-sh/ruff/pull/27584))
```
AlexWaygood: I'd classify the first and third of these as "Core type-checking", personally

AlexWaygood (approved):

.github/workflows/build-binaries.yml
```diff
@@ -450,6 +450,7 @@ jobs:
           - target: s390x-unknown-linux-gnu
             arch: s390x
             manylinux: 2_17
+            maturin_docker_options: -e CFLAGS_s390x_unknown_linux_gnu=-march=z10
```
charliermarsh, line 453: This is for Stacker.

Per Codex:
> The s390x release build sets `CFLAGS_s390x_unknown_linux_gnu=-march=z10` so the Stacker `psm` dependency can assemble its `lay` instructions. `z10` is already the default `rustc` CPU for `s390x-unknown-linux-gnu`, but the separate `s390x-ibm-linux-gnu-gcc` cross-compiler does not inherit that target CPU and otherwise defaults to an older instruction set. This aligns native C and assembly compilation with the existing Rust baseline without changing hardware compatibility or disabling Stacker.

PR #4281, opened 2026-08-16
Add dependency cooldowns

woodruffw:
## Summary

This is the same as with Ruff: we're adding 7-day cooldowns to every PEP 723 script.

## Test Plan

NFC.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ dev = ["prek==0.4.11"]
 release = ["rooster"]
 
 [tool.uv]
+exclude-newer = "P7D"
 cache-keys = [
     { file = "pyproject.toml" },
     { file = "dist-workspace.toml" },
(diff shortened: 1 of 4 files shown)


charliermarsh (approved):

pyproject.toml
```diff
@@ -44,6 +44,7 @@ dev = ["prek==0.4.11"]
 release = ["rooster"]
 
 [tool.uv]
+exclude-newer = "P7D"
```
AlexWaygood, line 47: (this isn't a PEP-723 script but I don't think we upgrade the lockfile automatically with renovate, so I think the same rationale applies for having a shorter cooldown period than for our renovate-updated deps)

PR #4283, opened 2026-08-16
Pin one-shot tools

woodruffw:
## Summary

Adds concrete version (and/or hash) pins where appropriate.

~~Note: I'm pretty sure the `ty` pins are over-eager here. But maybe Rooster can also auto-update these?~~

I think I did the Rooster changes correctly...

## Test Plan

NFC.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -131,4 +131,4 @@ jobs:
       - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
 
       - name: "Build docs"
-        run: uvx --with-requirements docs/requirements.txt mkdocs build --strict -f mkdocs.yml
+        run: uvx --from "mkdocs==1.6.0" --with-requirements docs/requirements.txt mkdocs build --strict -f mkdocs.yml
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,7 +103,7 @@ git add ruff
 To preview any changes to the documentation locally run the development server with:
 
 ```shell
-uvx --with-requirements docs/requirements.txt -- mkdocs serve -f mkdocs.yml
+uvx --from "mkdocs==1.6.0" --with-requirements docs/requirements.txt mkdocs serve -f mkdocs.yml
 ```
 
 The documentation should then be available locally at
@@ -122,7 +122,7 @@ Cloudflare Pages.
 After making changes to the documentation, format the markdown files with:
 
 ```shell
-npx prettier --prose-wrap always --write "**/*.md"
+npx prettier@3.9.6 --prose-wrap always --write "**/*.md"
 ```
 
 ## Releasing ty
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,18 @@ RUN case "$TARGETPLATFORM" in \
     esac
 # Update rustup whenever we bump the rust version
 COPY ruff/rust-toolchain.toml rust-toolchain.toml
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --target $(cat rust_target.txt) --profile minimal --default-toolchain none
+ARG BUILDARCH
+ARG RUSTUP_VERSION=1.28.2
+RUN case "$BUILDARCH" in \
+        "amd64") rustup_target="x86_64-unknown-linux-gnu"; rustup_sha256="20a06e644b0d9bd2fbdbfd52d42540bdde820ea7df86e92e533c073da0cdd43c" ;; \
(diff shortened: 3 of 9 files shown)


Dockerfile
```diff
@@ -21,7 +21,18 @@ RUN case "$TARGETPLATFORM" in \
...
+RUN case "$BUILDARCH" in \
+        "amd64") rustup_target="x86_64-unknown-linux-gnu"; rustup_sha256="20a06e644b0d9bd2fbdbfd52d42540bdde820ea7df86e92e533c073da0cdd43c" ;; \
+        "arm64") rustup_target="aarch64-unknown-linux-gnu"; rustup_sha256="e3853c5a252fca15252d07cb23a1bdd9377a8c6f3efa01531109281ae47f841c" ;; \
+        *) exit 1 ;; \
+    esac && \
+    curl --proto '=https' --tlsv1.2 -sSf "https://static.rust-lang.org/rustup/archive/${RUSTUP_VERSION}/${rustup_target}/rustup-init" -o /tmp/rustup-init && \
+    echo "${rustup_sha256}  /tmp/rustup-init" | sha256sum -c - && \
+    chmod +x /tmp/rustup-init && \
+    /tmp/rustup-init -y --target "$(cat rust_target.txt)" --profile minimal --default-toolchain none && \
+    rm /tmp/rustup-init
```
woodruffw, line 35: I can't say I love this...maybe there's a first-party Rust image we can hash-pin to instead? I'm not sure if that's better.

.github/workflows/ci.yaml
```diff
@@ -131,4 +131,4 @@ jobs:
       - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
 
       - name: "Build docs"
-        run: uvx --with-requirements docs/requirements.txt mkdocs build --strict -f mkdocs.yml
+        run: uvx --from "mkdocs==1.6.0" --with-requirements docs/requirements.txt mkdocs build --strict -f mkdocs.yml
```
AlexWaygood, line 134: I'd suggest pinning mkdocs as a dev-dependency in pyproject.toml and then changing this to `uv run --only-group=dev --locked`

CONTRIBUTING.md
```diff
@@ -103,7 +103,7 @@ git add ruff
 To preview any changes to the documentation locally run the development server with:
 
 ```shell
-uvx --with-requirements docs/requirements.txt -- mkdocs serve -f mkdocs.yml
+uvx --from "mkdocs==1.6.0" --with-requirements docs/requirements.txt mkdocs serve -f mkdocs.yml
```
AlexWaygood, line 106: Same here

Dockerfile
AlexWaygood, line 1: I cannot review this file, I know nothing about docker 😆

docs/reference/typing-faq.md
```diff
@@ -434,7 +434,7 @@ it with ty by using uv's `--with-requirements` flag to install the dependencies
 script header:
 
 ```bash
-uvx --with-requirements script.py ty check script.py
+uvx --with-requirements script.py ty@0.0.72 check script.py
```
AlexWaygood, line 437: We might be able to get rid of this uvx invocation altogether and replace it with `uv check` either now or soon (cc @MichaReiser). But that definitely doesn't need to be done in this PR

.github/workflows/ci.yaml
AlexWaygood: Looking more closely, possibly we should convert the whole `docs/requirements.txt` file into a `docs` dependency group in pyproject.toml

AlexWaygood (approved):
I don't think any of my comments should block merging this on a second look, they can all be followups

Dockerfile
zanieb: We should make this change separately, it's way more sensitive than pinning one-off uvx invocations.

zanieb: Also, we should match across our repositories, e.g., this file is already drifting from uv https://github.com/astral-sh/uv/blob/main/Dockerfile

AlexWaygood:
Seems like the best way forward here is to rebase this and split out the Docker changes into a standalone PR?

.github/workflows/ci.yaml
mgaitan: my pattern for this is having the dev group defined as  tests + docs + lint groups.  Then you can use `--group docs|test|...`  depending the workflow

https://mgaitan.github.io/python-package-copier-template/design_decisions.html#dependency-management-with-uv

PR #4314, opened 2026-08-18
Harden the Prettier hook with `prek`

zaniebot:
The frozen `prettier@3.9.6` hook still depends on an implicit Node/npm installation policy. Keep the existing mirror hook, require `prek==0.4.13`, and let `prek` manage Node `24.19.0` with Renovate updates limited to Node 24. Configure `prek update` to freeze revisions and wait seven days for new tags. Set `npm_config_ignore_scripts=true` and `npm_config_min_release_age=7` in CI, and document the same caller-level settings for contributors. The YAML selection and formatting behavior remain unchanged. This freezes hook revisions, not the full transitive npm dependency tree.

Related: astral-sh/ty#4283 handles the documentation’s one-off tool pins. This leaves that command untouched to avoid a competing edit; moving it to `prek exec prettier` remains a separate follow-up. astral-sh/ruff#27842 applies the matching native `prek` policy, while astral-sh/ecosystem-analyzer#162 removes an unnecessary `prettier` hook.


CONTRIBUTING.md
```diff
@@ -46,10 +46,15 @@ git submodule update --init --recursive
...
-when making a commit:
+`prek` manages the pinned Node.js version used by JavaScript hooks. Set the following npm options in
+your shell before running `prek` or committing with installed hooks. They disable lifecycle scripts
+and require newly resolved npm versions to be at least seven days old.
+
+You can optionally install prek hooks to automatically run the validation checks when making a commit:
 
 ```shell
+export npm_config_ignore_scripts=true
+export npm_config_min_release_age=7
```
zanieb, line 57: Can we encode these in the pre-commit config instead?

j178: These could be configured globally via [`default_env`](https://prek.j178.dev/reference/configuration/#default_env), but they are only applied during hook environment installation starting with prek [0.5.1](https://github.com/j178/prek/releases/tag/v0.5.1). In earlier releases, they only took effect at hook runtime.

PR #4315, opened 2026-08-18
Manage docs build with a dep group

woodruffw:
## Summary

This switches us to a proper dep group (named `docs`) for the MkDocs based doc build.

Also addresses a bunch of Dependabot advisories in the process.

## Test Plan

Ran it locally, we'll see what happens in CI.


pyproject.toml
```diff
@@ -42,6 +42,17 @@ build-backend = "maturin"
 [dependency-groups]
 dev = ["prek==0.4.11"]
 release = ["rooster"]
+docs = [
+    "black>=23.10.0",
```
AlexWaygood: Our docs build depends on black?!

AlexWaygood (approved):

pyproject.toml
woodruffw: 🤔 this was in the `requirements.in`. I'll track it down.

woodruffw: Removed and nothing broke, so I think it's okay to just delete...

AlexWaygood: it was probably copied-and-pasted from Ruff before we removed Ruff's dependency on black

AlexWaygood (approved):

PR #4319, opened 2026-08-18
Harden release workflow permissions and shell inputs

zaniebot:
The release workflow grants `contents: write` by default and embeds release tags in shell commands. Restrict the default token, declare the read access needed by planning, building, and reusable workflows, and preserve the existing publishing permissions. Pass release values through environment variables and quoted argument arrays so they are treated as arguments rather than shell source.

Port of astral-sh/uv#19822.

PR #4320, opened 2026-08-18
Document manual maintenance of the release workflow

zaniebot:
The release workflow is maintained manually, but its header and `AGENTS.md` still describe it as generated. Update the header and remove the obsolete agent guidance. `cargo-dist` remains responsible for release planning and artifact generation.

Follow-up to [astral-sh/ty#4319](https://github.com/astral-sh/ty/pull/4319#issuecomment-5332497565).


AGENTS.md
```diff
@@ -2,4 +2,4 @@
 
 ## Development
 
-Parts of `.github/workflows/release.yml` are generated by cargo-dist from `dist-workspace.toml`. Before editing the release workflow, check whether the relevant section is generated. Prefer changing `dist-workspace.toml` or the referenced reusable workflow instead of editing generated YAML. After modifying cargo-dist configuration, regenerate the workflow with the cargo-dist version pinned in `dist-workspace.toml` and inspect the resulting diff to ensure the change will survive future regenerations.
+`.github/workflows/release.yml` is maintained manually. Edit it or the referenced reusable workflows directly. `cargo-dist` still uses `dist-workspace.toml` to plan releases and build artifacts, but regenerating the workflow can overwrite manual customizations. If you regenerate it, review the diff carefully and preserve those customizations.
```
zanieb: I think we could consider just removing this note entirely now

MichaReiser (approved):
Does that mean we won't use cargo dist anymore at all? Just trying to understand what I'm supposed to do in the future.

AlexWaygood:
Similar to ruff, we should consider applying our regular workflow linters on the file if it's no longer generated — the previous rationale for excluding it doesn't apply anymore https://github.com/astral-sh/ty/blob/4fcb06010f2219b80adf3e688964abd9b0d72866/.pre-commit-config.yaml#L5

PR #4338, opened 2026-08-19
Publish `ty_wasm` to npm

### LINKED ISSUE #4007: NPM WASM package
### Question

Hi there,

Are there any plans to publish `ty` on NPM as a wasm package, similar to how you did it for [ruff](https://www.npmjs.com/package/@astral-sh/ruff-wasm-web)?

Thanks
Tobias

### Version

_No response_

typedrat:
## Summary

Closes #4007.

Adds a `build-wasm` local-artifacts job and a `publish-wasm` publish job to the `dist` pipeline, publishing `ty_wasm` as `@astral-sh/ty-wasm-web`, `@astral-sh/ty-wasm-bundler` and `@astral-sh/ty-wasm-nodejs`. Both workflows are ports of ruff's `build-wasm.yml` and `publish-wasm.yml`.

The build side needed little. `ty_wasm` is already built with `wasm-pack` on every push to `main` for the playground, already lints under `--target wasm32-unknown-unknown`, and already runs `wasm-pack test --node` in ruff's CI.

Versioning is the one place this can't follow ruff. `ruff_wasm/Cargo.toml` carries the real version and is bumped in-repo, so `wasm-pack` writes it into `package.json`. `ty_wasm` is pinned at `0.0.0`, ty's version exists only in this repository, and `scripts/release.sh` aborts on a dirty submodule, so rooster can't write to `ruff/crates/ty_wasm/Cargo.toml`. The build job instead reads the version from `pyproject.toml`, where the wheels get theirs, and injects it into the generated `package.json` along with the name. The same step corrects `homepage`, which `ty_wasm` inherits from the ruff workspace and which would otherwise point the packages at ruff's documentation. Taking it from `announcement_tag` in the dist plan, or bumping `ty_wasm`'s `Cargo.toml` upstream, would work too if you'd prefer either.

Three other deviations from ruff:

- `rustup target add` runs with `working-directory: ruff`. The toolchain pin lives in the submodule rather than at this repository's root, so at the root the target lands on the runner's default toolchain and `wasm-pack` then fails to build.
- The build runs on `depot-ubuntu-24.04-4`. `ty_wasm` links the whole type checker and the vendored typeshed under `lto = "fat"`, so it's a good deal heavier than `ruff_wasm`.
- `publish-wasm` gets `contents: read` and `id-token: write`, without ruff's `packages: write`. `npm publish` doesn't touch GitHub Packages, but I may be missing something.

Trusted publishers still need setting up on npmjs for the three names, which are unregistered today. Until that happens the publish job will fail on a real release; dry-runs are unaffected.

For size reference, each package comes out at 7.1 MB packed and 17.1 MB unpacked, against 11.05 MB unpacked for the published `@astral-sh/ruff-wasm-web`.

One omission: `ty_wasm` has no `README.md`, so the npm pages would render empty. That belongs beside the crate in `astral-sh/ruff` rather than here, so it isn't in this PR. I can open it separately.

## Test Plan

`release.yml` is hand-maintained (`allow-dirty = ["ci"]`), so I didn't regenerate it. I generated a clean copy from the updated `dist-workspace.toml` with dist 0.31.0 and diffed it against this branch. The only differences are the hand-patches already present, plus the same three that the sibling jobs already carry applied to the new ones: `$/` in `uses:`, explicit `contents: read` on the build job, and `release-gate` in the publish job's `needs`. Both new jobs otherwise match ruff's `custom-build-wasm` and `custom-publish-wasm`, apart from the permissions noted above.

`prek run` (zizmor, prettier, check-github-workflows) and `prek run actionlint --hook-stage=manual` pass, and `dist plan` accepts the updated config. The `jq` rename step was run against the real `package.json` from the published `@astral-sh/ruff-wasm-web` tarball; it replaces name and version and leaves every other field alone.

I ran both workflows end to end in a container approximating the ubuntu-24.04 runner (rustup with a preinstalled stable toolchain, wasm-pack v0.13.1, Node 24), driving them with the `run:` bodies extracted from the YAML. All three targets build and `npm publish --dry-run` cleanly, the version resolves to 0.0.73, and `LICENSE` lands in the tarball. Two limits on that: the container approximates the runner rather than being it, and the three targets ran sequentially in one workspace, so cargo's cache was warm after the first. The publish path past `--dry-run` is unexercised.

---

Disclaimer: this was put together with AI assistance (Claude Code).

### DIFF
--- a/.github/workflows/build-wasm.yml
+++ b/.github/workflows/build-wasm.yml
@@ -44,6 +44,9 @@ jobs:
           submodules: recursive
           persist-credentials: false
       - name: "Install Rust toolchain"
+        # Run from `ruff`, so the target is added to the toolchain that
+        # `ruff/rust-toolchain.toml` pins rather than to the runner's default.
+        working-directory: ruff
         run: rustup target add wasm32-unknown-unknown
       - uses: jetli/wasm-pack-action@0d096b08b4e5a7de8c28de67e11e945404e9eefa # v0.4.0
         with:
@@ -53,11 +56,12 @@ jobs:
         run: wasm-pack build --target ${{ matrix.target }} ruff/crates/ty_wasm
       - name: "Rename generated package"
         # `ty_wasm` is pinned at version 0.0.0 in the submodule, since ty's version only exists in
-        # this repository. Take it from the same place the wheels do.
+        # this repository. Take it from the same place the wheels do. `homepage` is inherited from
+        # the ruff workspace and would otherwise point at ruff's docs.
         run: |
           version="$(python3 -c 'import pathlib, tomllib; print(tomllib.loads(pathlib.Path("pyproject.toml").read_text())["project"]["version"])')"
           jq --arg name "@astral-sh/ty-wasm-${{ matrix.target }}" --arg version "$version" \
-            '.name = $name | .version = $version' \
+            '.name = $name | .version = $version | .homepage = "https://docs.astral.sh/ty"' \
             ruff/crates/ty_wasm/pkg/package.json > /tmp/package.json
           mv /tmp/package.json ruff/crates/ty_wasm/pkg/package.json
       - run: cp LICENSE ruff/crates/ty_wasm/pkg # wasm-pack does not put the LICENSE file in the pkg
--- a/.github/workflows/build-wasm.yml
+++ b/.github/workflows/build-wasm.yml
@@ -1,0 +1,69 @@
+# Build ty_wasm for npm.
+#
+# Assumed to run as a subworkflow of .github/workflows/release.yml; specifically, as a local
+# artifacts job within `cargo-dist`.
+name: "Build wasm"
+
+on:
+  workflow_call:
+    inputs:
+      plan:
+        required: true
+        type: string
+  pull_request:
+    paths:
+      - .github/workflows/build-wasm.yml
+
+concurrency:
+  group: build-wasm-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+permissions:
+  contents: read
+
+env:
+  ACTIONS_CACHE_MODE: none
+  CARGO_INCREMENTAL: 0
+  CARGO_NET_RETRY: 10
+  CARGO_TERM_COLOR: always
+  RUSTUP_MAX_RETRIES: 10
+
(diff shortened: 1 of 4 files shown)


typedrat:
I ran both workflows end to end in a container approximating the ubuntu-24.04 runner (rustup with a preinstalled stable toolchain, wasm-pack v0.13.1, Node 24), driving them with the `run:` bodies extracted from the YAML. It caught a real bug, fixed in e0d7211.

`rustup target add` ran at the repository root, where no `rust-toolchain.toml` is in scope, so wasm32 went to the default stable toolchain. `wasm-pack` builds from `ruff/crates/ty_wasm`, where 1.97.1 is pinned, and that toolchain had no wasm32 std:

```
error[E0463]: can't find crate for `std`
  = note: the `wasm32-unknown-unknown` target may not be installed
```

ruff avoids this because its `rust-toolchain.toml` is at its repository root. The fix is `working-directory: ruff`, matching `ci.yaml` and `build-binaries.yml`.

Separately, `ty_wasm` inherits `homepage` from the ruff workspace, so the packages advertised ruff's docs; the rename step now overrides it to `https://docs.astral.sh/ty`. Setting it in `ty_wasm`'s manifest upstream would be tidier if you'd prefer that. `repository` still points at `astral-sh/ruff`, where the crate actually lives.

All three targets now build and `npm publish --dry-run` cleanly at 7.1 MB packed / 17.1 MB unpacked, the version resolves to 0.0.73, `LICENSE` is in the tarball, and the wasm matches the playground artifact's size.

Caveats: the container approximates the runner rather than being it, and the three targets ran sequentially in one workspace, so cargo's cache was warm after the first. The publish path past `--dry-run` is still unexercised, and still needs trusted publishers for the three names.

MichaReiser:
Thank you. I plan to take a look at this. We'll need to publish a placeholder package before merging, restrict publishing. 

Codex also mentioned something that we need to fix the package's url. I have to take a closer look at what it meant by this

typedrat:
Hey, just want to check in on this.

MichaReiser:
Yeah sorry. This requires some setup on our end as well (setting up the project etc), which I just haven't found the time to do so yet

typedrat:
no worries, just wanted to make sure there wasn't anything I could do to help 🙂

PR #4339, opened 2026-08-20
Document uv integration and workspace trust settings

MichaReiser:
## Summary

Document `experimental.useUv` and `untrustedWorkspace`, including their defaults, behavior, and editor examples.
Companion to https://github.com/astral-sh/ty-vscode/pull/517 and https://github.com/astral-sh/ruff/pull/27619.

## Test Plan

Testing: Markdown checks and the strict documentation build passed.

### DIFF
--- a/docs/reference/editor-settings.md
+++ b/docs/reference/editor-settings.md
@@ -665,12 +665,13 @@ ______________________________________________________________________
 
 ## Initialization options
 
-The following options are read when ty is initialized in an editor. Changing them requires
-restarting the language server.
+The following settings are required when ty is initialized in an editor. These settings are
+static so changing them requires restarting the editor to take effect.
 
-In VS Code, most of these options are available in the `ty.*` namespace. The extension sets
-[`untrustedWorkspace`](#untrustedworkspace) automatically. In other editors, use the field for
-language server initialization options. Refer to the examples below.
+For VS Code users, most of these settings are defined in the `ty.*` namespace as usual, but for other
+editors, they would need to be provided in a separate field of the configuration that corresponds to
+the initialization options. Refer to the examples below for how to set these options in different
+editors.
 
 ### `experimental.useUv`
 
@@ -854,7 +855,6 @@ external commands. This disables uv integrations and prevents ty from running Py
 to discover environment information.
 
 Set this option in trusted editor configuration before opening a workspace you do not trust.
-Changing it requires restarting the language server.
 
 **Default value**: `false`
 
--- a/docs/reference/editor-settings.md
+++ b/docs/reference/editor-settings.md
@@ -683,12 +683,9 @@ Control how ty uses [uv](https://docs.astral.sh/uv/):
 - `on`: Use uv for project discovery and standalone script environments.
 
 This feature is experimental and may change. Enabling it requires
-[uv to be installed](https://docs.astral.sh/uv/getting-started/installation/) and can download and
+[uv 0.12.3 or later](https://docs.astral.sh/uv/getting-started/installation/) and can download and
 install dependencies. All uv integrations are disabled when
 [`untrustedWorkspace`](#untrustedworkspace) is `true`.
-
-If this option is not provided, the language server uses the `TY_UV` environment variable.
-The VS Code extension sends `"off"` by default.
 
 **Default value**: `"off"`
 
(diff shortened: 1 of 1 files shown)


docs/reference/editor-settings.md
```diff
@@ -668,11 +668,79 @@ ______________________________________________________________________
...
+    {
+      "lsp": {
+        "ty": {
+          "initialization_options": {
+            "experimental": {
+              "useUv": "scripts"
+            }
+          }
+        }
+      }
```
SecT0uch, line 737: This is not working for me.

```
uv 0.12.7 (61291a8ca 2026-08-27 x86_64-unknown-linux-gnu)
# ty provided by Zed
ty Version: 0.0.75
Zed 1.17.2
```

SecT0uch: It now works with the last version! Thanks

docs/reference/editor-settings.md
```diff
@@ -668,11 +668,79 @@ ______________________________________________________________________
...
+Control how ty uses [uv](https://docs.astral.sh/uv/):
+
+- `off`: Do not use uv.
+- `scripts`: Use uv to create and update environments for standalone scripts with
+    [PEP 723 inline metadata](https://peps.python.org/pep-0723/).
+- `on`: Use uv for project discovery and standalone script environments.
+
+This feature is experimental and may change. Enabling it requires
+[uv 0.12.3 or later](https://docs.astral.sh/uv/getting-started/installation/) and can download and
+install dependencies. All uv integrations are disabled when
```
dhruvmanila: Does the version requirement satisfy the "can download and install dependencies"? If so, I'd remove that wording to make it simply that it requires a certain uv version.

docs/reference/editor-settings.md
```diff
@@ -668,11 +668,79 @@ ______________________________________________________________________
...
+
+    ```lua
+    -- Neovim >=0.11:
+    vim.lsp.config('ty', {
+      init_options = {
+        experimental = {
+          useUv = 'scripts',
+        },
+      },
+    })
```
dhruvmanila, line 713: `experimental` is part of the global options, so I think we should recommend using `settings` instead:


```suggestion
    vim.lsp.config('ty', {
      settings = {
        ty = {
          experimental = {
            useUv = 'scripts',
          },
        },
      },
    })
```

same needs to be updated for other editors as well

docs/reference/editor-settings.md
```diff
@@ -775,4 +843,56 @@ The log level to use for the language server.
...
+
+=== "Neovim"
+
+    ```lua
+    -- Neovim >=0.11:
+    vim.lsp.config('ty', {
+      init_options = {
+        untrustedWorkspace = true,
+      },
+    })
```
dhruvmanila, line 873: This reads like a workspace specific option but is applied globally and is not specific to any workspace. But, I think for VS Code it is on a per-workspace level given that it occurs automatically. Neovim doesn't have multiple workspaces but I think Zed does, do we have if this creates any discrepancy in behavior between editors?

Sorry to highlight this right now, it might not be an issue at all.

dhruvmanila (approved):

docs/reference/editor-settings.md
MichaReiser: Hmm, not sure. It's currently only supported as part of the initialization options (this may change later)

docs/reference/editor-settings.md
MichaReiser: We also pick the ty executable based on the global trust, and not on the trust level per workspace folder. I don't think we should make a distinction here. 

I also tested this in VS Code. Opening two workspace folders and trusting the workspace results in VS Code trusting **all** open workspace folders. So this setting is global and not per workspace folder.

docs/reference/editor-settings.md
MichaReiser:
```suggestion
This feature is experimental and may change. Enabling it requires
[uv 0.12.3 or later](https://docs.astral.sh/uv/getting-started/installation/). All uv integrations are disabled when
```

docs/reference/editor-settings.md
dhruvmanila: Got it, thanks for checking

docs/reference/editor-settings.md
dhruvmanila: Oh I see, yeah then this makes sense, didn't notice that `useUv` is being discarded post initialization

MichaReiser: Yeah, I think we should change this behavior and have it be accepted in the `globalOptions`, but for now, this is how it works.

PR #4384, opened 2026-08-25
Fix documented Python package build command

### LINKED ISSUE #4361: `Building source distributions is disabled` error when building local Python package
### Summary

`CONTRIBUTING.md` states that a ty Python package can be built with `uv build`, but that no longer works after commit b25dd98876:

```
$ uv build
Building source distribution...
error: Failed to build `/home/user/ty`
  Caused by: Building source distributions is disabled

$ uv --version
uv 0.12.5 (210d1f678 2026-08-14 x86_64-unknown-linux-gnu)
```

### Version

00199f0aaa6a

carljm:
## Summary

Use `uv build --build` to override the repository's `no-build = true` setting when contributors build the Python package.

Fixes #4361.

## Test plan

Documentation-only change; no new tests. Verified that this works to build ty.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,7 +58,7 @@ uv run --only-group dev --locked prek install
 The Python package can be built with any Python build frontend (Maturin is used as a backend), e.g.:
 
 ```shell
-uv build
+uv build --build
 ```
 
 ## Updating the Ruff commit


zanieb:
You can also allow-list ty via `no-binary-package = ["ty"]`

zanieb:
(which is probably a fine solution here)

zanieb:
We should also allow first-party builds in `uv build` even with `no-build`. I guess I'll fix that too.

zanieb:
This looks vaguely intentional in the code so we'll need to discuss the change.

zanieb (approved):

carljm:
@zanieb it seems like `no-binary-package = ["ty"]` does not fix the issue; it still produces "Building source distributions is disabled" on `uv build`. Codex says:

> uv’s [[build frontend](https://github.com/astral-sh/uv/blob/0.12.5/crates/uv/src/commands/build_frontend.rs#L760)](https://github.com/astral-sh/uv/blob/0.12.5/crates/uv/src/commands/build_frontend.rs#L760) passes no package identity, so the exception cannot match.

So I'll go ahead with this docs change for now.

PR #4398, opened 2026-08-26
Improve `uv-lock` pre-commit configuration

AlexWaygood:
## Summary

Same as https://github.com/astral-sh/ruff/pull/28078. `--check` gives a better error message when the hook fails in CI; `--refresh` means the hook will reject manual updates to the lockfile that don't abide by the cooldown policy.

## Test Plan

`uv run --locked --only-group=dev prek run --all-files`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,7 @@ repos:
     hooks:
       - id: uv-lock
         priority: 0
+        args: [--check, --refresh, --default-index=https://pypi.org/simple]
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: 3e8a8703264a2f4a69428a0aa4dcb512790b2c8c # frozen: v6.0.0


zanieb:
hm why `--refresh`? That shouldn't be necessary for the behavior described?

AlexWaygood:
> hm why `--refresh`? That shouldn't be necessary for the behavior described?

That was not what I observed in https://github.com/astral-sh/ecosystem-analyzer/pull/168.

The PR was made by renovate, and upgraded lockfile pins to some recent releases. I then added a uv cooldown period in https://github.com/astral-sh/ecosystem-analyzer/pull/171 and the `uv-lock` prek hook in https://github.com/astral-sh/ecosystem-analyzer/pull/172, and rebased https://github.com/astral-sh/ecosystem-analyzer/pull/168 on `main` after merging those PRs, but CI still passed on https://github.com/astral-sh/ecosystem-analyzer/pull/168 despite the fact that it was upgrading lockfile pins to versions that had been released inside the cooldown period. It was only after I added `--refresh` in https://github.com/astral-sh/ecosystem-analyzer/pull/173 and rebased the renovate PR on top of _that_ that CI started to fail on the renovate PR.

PR #4439, opened 2026-09-01
fix: update broken registry.bazel.build/docs URLs

georgesfarah:
## Summary

The `registry.bazel.build/docs/<name>` URL pattern returns 404 and is no longer valid.

The correct URL format is now `registry.bazel.build/modules/<name>/latest/docs`.

This was identified across multiple Bazel rule repositories — see https://github.com/hermeticbuild/rules_rs/pull/235 for the original fix.

## Changes

- Replaced broken `/docs/aspect_rules_lint` URL with `/modules/aspect_rules_lint/latest/docs`
- Replaced symbol anchor with explicit path:
  - `#function-lint_ty_aspect` → `/lint/ty.bzl/lint_ty_aspect`

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -169,7 +169,7 @@ The following tags are available:
 
 ### Using ty with Bazel
 
-[`aspect_rules_lint`](https://registry.bazel.build/docs/aspect_rules_lint#function-lint_ty_aspect)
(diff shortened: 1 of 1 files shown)


sharkdp (approved):
Thank you

PR #4457, opened 2026-09-03
Allow Ruff to format our docs examples

AlexWaygood:
## Summary

It feels like a bit of a bad sign for the feature if we can't enable Ruff's markdown formatting on some of our own repos. But I think we can -- most of the formatting changes Ruff wants to make here seem like clear improvements to me. The one clear exception is in `suppression.md`, but there it's fairly easy to suppress Ruff's formatting with an HTML comment that will be invisible to readers of the docs.

## Test Plan

`uvx prek run -a` passes

### DIFF
--- a/docs/features/type-system.md
+++ b/docs/features/type-system.md
@@ -30,10 +30,7 @@ def output_as_json(obj: Serializable) -> str:
     if isinstance(obj, Versioned):
         reveal_type(obj)  # reveals: Serializable & Versioned
 
-        return str({
-            "data": obj.serialize_json(),
-            "version": obj.version
-        })
+        return str({"data": obj.serialize_json(), "version": obj.version})
     else:
         return obj.serialize_json()
 ```
@@ -71,8 +68,10 @@ protocol, accounting for the possibility of subclasses of `Animal` that add a `n
 class Person:
     name: str
 
+
 class Animal:
     species: str
+
 
 def greet(being: Person | Animal | None):
     if hasattr(being, "name"):
@@ -103,8 +102,9 @@ if TYPE_CHECKING:
 
     type SerializableVersioned = Intersection[Serializable, Versioned]
 
+
 def output_as_json(obj: SerializableVersioned) -> str:
-    ...
+    raise NotImplementedError
 ```
 
 (Full example in the [playground](https://play.ty.dev/f003e901-0e45-4f45-9759-d6db9d5e5f66))
@@ -123,6 +123,7 @@ of `list[Unknown]`:
 ```py
 @final
 class Item: ...
+
 
 def process(items: Item | list[Item]):
     if isinstance(items, list):
@@ -158,8 +159,10 @@ from pydantic import BaseModel
 
 PYDANTIC_V2 = pydantic.__version__.startswith("2.")
 
+
 class Person(BaseModel):
     name: str
+
 
 def to_json(person: Person):
     if PYDANTIC_V2:
--- a/docs/features/type-system.md
+++ b/docs/features/type-system.md
@@ -103,8 +103,7 @@ if TYPE_CHECKING:
     type SerializableVersioned = Intersection[Serializable, Versioned]
 
 
-def output_as_json(obj: SerializableVersioned) -> str:
-    raise NotImplementedError
+def output_as_json(obj: SerializableVersioned) -> str: ...
 ```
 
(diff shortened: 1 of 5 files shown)


docs/features/type-system.md
```diff
@@ -30,10 +30,7 @@ def output_as_json(obj: Serializable) -> str:
     if isinstance(obj, Versioned):
         reveal_type(obj)  # reveals: Serializable & Versioned
 
-        return str({
-            "data": obj.serialize_json(),
-            "version": obj.version
-        })
```
sharkdp, line 36: I liked the previous version better, which is why I chose to not format the docs at all, but I certainly don't feel too strongly :D

docs/features/type-system.md
```diff
@@ -103,8 +102,9 @@ if TYPE_CHECKING:
 
     type SerializableVersioned = Intersection[Serializable, Versioned]
 
+
 def output_as_json(obj: SerializableVersioned) -> str:
-    ...
+    raise NotImplementedError
```
sharkdp: I think this raises (heh) questions. I liked the previous version, but if you want to go through with this, I'd prefer a trailing `...` over this.

docs/reference/typing-faq.md
```diff
@@ -318,7 +333,7 @@ the developer experience around this in the future.
 
 
     def retry(times: int, operation: FunctionLikeCallable[[], bool]) -> bool:
-        ...
+        raise NotImplementedError
```
sharkdp: same here

docs/reference/typing-faq.md
```diff
@@ -143,10 +147,12 @@ The starred spellings only appear in ty's output; they cannot be used in Python
     else:
         JustFloat = float
 
+
     def only_actual_floats_allowed(f: JustFloat) -> None: ...
 
+
     only_actual_floats_allowed(1.0)  # OK
-    only_actual_floats_allowed(1)    # error: invalid-argument-type
+    only_actual_floats_allowed(1)  # error: invalid-argument-type
```
sharkdp, line 151: also not an improvement

docs/reference/typing-faq.md
```diff
@@ -69,8 +69,11 @@ produces a new type, ty replaces the non-convergent part with `Divergent`.
 For example, each iteration of this loop wraps `x` in another list:
 
 ```py
+import random
+
+
 def some_condition() -> bool:
-    ...
+    return random.choice([True, False])
```
sharkdp: this seems like unnecessary noise that distracts from the point we want to make.

sharkdp (commented):
> It feels like a bit of a bad sign for the feature if we can't enable Ruff's markdown formatting on some of our own repos.

I don't agree with this argumentation. Docs are just different. Docs don't contain real code, they contain educational examples that were written in a very specific way.

I don't think this PR is an improvement, but I won't object if you want to follow through with it.

docs/features/type-system.md
AlexWaygood: I worry that it's confusing to users if our docs examples cause ty to emit spurious errors. It means that they can't copy and paste them into the playground. Here, for example, the way the function is on `main`, we'd complain that the function is annotated as returning `str`, but actually returns `None` at runtime.

This is orthogonal to the PR's main objective, though, so I'll just revert this change for now

docs/features/type-system.md
AlexWaygood: hehe, well, the solution for avoiding arguments over which formatting style is better is to introduce an autoformatter ;-)

docs/reference/typing-faq.md
AlexWaygood: I agree, but this feels like something that should be brought up on the Ruff issue tracker (I think @MichaReiser has argued for a long time that black and Ruff are too picky about the number of spaces they allow prior to comments, for example). Dogfooding ensures we feel the pain of enforcing our autoformatter ;)

sharkdp: > Dogfooding ensures we feel the pain of enforcing our autoformatter ;)

We should feel the pain, somewhere, sure. But users should not feel pain when reading our docs. I'll let you decide, I'm not going to engage in any of the discussions here going forward. I don't think this is important enough to require two of us to decide.

MichaReiser: I argue that black and ruff are too opinionated when it comes to most whitespace handling 😆

PR #4488, opened 2026-09-08
Fix code block syntax highlighting in rules.md

rzwq:
The language specifier has been fixed to ensure correct code display (some extensions in VS Code for `.md` do not support `py`, only `python`).

### DIFF
--- a/docs/reference/rules.md
+++ b/docs/reference/rules.md
@@ -110,7 +110,7 @@ class are explicitly declared.
 **Examples**
 
 
-```py
+```python
 from typing import ClassVar, Protocol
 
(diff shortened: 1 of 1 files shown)


MichaReiser:
Thank you, but this file is auto generated. You have to make the change in the https://github.com/astral-sh/ruff repository instead. I'm also not convinced this is worth changing in our docs. `py` and `python` are well established language identifiers and it may not be possible to write our docs in a way that they're supported by all extensions (what if another extension only supports `python` but not `py`?). But we can discuss this if this is a very popular extensin.

PR #4498, opened 2026-09-09
Remove redundant --exit-non-zero-on-fix from Ruff hook

ooooo-create:
## Summary

Remove `--exit-non-zero-on-fix` from the Ruff hook, keeping `--fix`. prek already fails when a hook modifies files.
This follows https://github.com/astral-sh/ruff-pre-commit/pull/57.
## Test Plan

- `uv run --only-group dev --locked prek validate-config .pre-commit-config.yaml`
- `uv run --only-group dev --locked prek run --files .pre-commit-config.yaml`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -106,7 +106,7 @@ repos:
     rev: 321478e58f4938179c6b86e4ddfa923d1547a49b # frozen: v0.16.6
     hooks:
       - id: ruff
-        args: [--fix, --exit-non-zero-on-fix]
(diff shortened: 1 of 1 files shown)


MichaReiser:
Same as for the same change in Ruff. I see some value in specifying the option

> Thank you. I think it makes sense to keep `--exit-non-zero-on-fix`. It has the benefit that the ruff hook fails. Without it, prek only reports that the group shared with markdownlint-fix failed. Therefore, I don't think the flag is entirely redundant.

PR #4537, opened 2026-09-16
Remove accidental uv version constraint

charliermarsh:
Remove the uv version constraint accidentally added in #4501. This allows us to update the uv pre-commit hook without a conflicting project-level pin.


AlexWaygood (approved):
(curious how this was added accidentally 😄)

charliermarsh:
Codex

charliermarsh:
> I added it manually in release PR #4501 as a CI workaround. setup-uv was selecting uv 0.12.12, while the lockfile had been generated with 0.12.11; changed handling of exclude-newer = "P7D" made both uv run --locked and the generated-files check report that the lock needed updating. Since setup-uv reads [tool.uv].required-version, pinning 0.12.11 made CI deterministic.

> But it was too broad and poorly scoped. It pinned uv for every contributor and workflow, wasn’t generated by the release script, and should not have been slipped into a release PR without explanation. The 0.0.81 PR only inherited it. The current checkout already has commit 93e8567 removing it.

MichaReiser:
Thank you

PR #4581, opened 2026-09-23
publish to astral-sh/versions via OIDC

ewdurbin:
adopt OIDC for release publication to the astral-sh/versions using OIDC rather than a long lived PAT. similar to docs publishing

Policies merged/applied in https://github.com/astral-sh/github-policies/pull/78
Versions STS merged in https://github.com/astral-sh/versions/pull/194
`RELEASES_STS_API_URL` configured in `release` environment.


.github/workflows/publish-versions.yml
```diff
@@ -31,8 +33,25 @@ jobs:
       - name: "Set branch name"
         run: echo "BRANCH_NAME=update-versions-$VERSION-$(date +%s)" >> "$GITHUB_ENV"
 
-      - name: "Clone versions repo"
-        run: git clone https://${{ secrets.ASTRAL_VERSIONS_PAT }}@github.com/astral-sh/versions.git astral-versions
```
woodruffw, line 35: Same Q as in uv, I assume this is safe to remove now.

woodruffw (approved):

.github/workflows/publish-versions.yml
ewdurbin: yes, post-merge we can remove the PAT, and ideally revoke it entirely. there will also be a followup PR to update the policies to remove that user's access to the versions repo.

